PR #7, opened 2020-03-31
Cleanup the buffer type checking function in the CUDA backend

pavanbalaji:
## Pull Request Description

The buffer type checking function was defined as a static inside the device exposed header file.  This was causing it to be included in files that didn't use the function at all and was throwing warnings.  We could have silenced it with an attribute, but it is cleaner to simply move the function into a `.c` file.

## Expected Impact

Fixes warnings.

## Author Checklist
* [x] Passes all tests


pavanbalaji:
Rebased on to the latest master to linearize commit history.

PR #10, opened 2020-03-31
build: update testing

hzhou:
## Pull Request Description

* Update `runtests.py` 
* Remove sub configure for test
* Consolidate test Makefiles into the main Makefile to allow fully dependency tracking

## Background

The testing program, `pack`, `iov`, etc. get built during `make [all]` and they are not updated when `libyaksa.la` is modified.

The main goal is to ensure when a yaksa source get updated, `make testing` inside a subdirectory will trigger the full remake, including `libyaksa.la` and test programs.

## Author Checklist
* [x] Passes all tests


hzhou:
test:yaksa

hzhou:
test:yaksa

hzhou:
test:yaksa

hzhou:
test:yaksa

hzhou:
@pavanbalaji Ready for your review.

hzhou:
test: yaksa

hzhou:
test: yaksa

hzhou:
> Also, is this fixing the primary issue of dependency between the test files and the yaksa source files? When I tried it out, it didn't seem to do that.

If you run `make testing` in the top dir, since the `testing` target depends on `libyaksa.la`, and `libyaksa.la` depends on its sources, so I believe it should work -- could you verify?

Under the `test` subfolders, since they are using separate Makefiles, the source dependency on `libyaksa.la` is not visible. It will remake the targets if `libyaksa.la` gets updated, but it won't happen automatically if its sources are changed. As a remedy, we could add `$(MAKE) -C $(top_srcdir) all` into the rules of `testing` target. Then it will always remake `libyaksa.la` if its sources updated, which will trigger the remake of `pack` etc. If we do this, we could revert to use `nodist_PROGRAM`.

hzhou:
> As a remedy, we could add `$(MAKE) -C $(top_srcdir) all` into the rules of `testing` target

Actually, it is probably better to add that as the make rule for the `libyaksa.la` target.

hzhou:
> > As a remedy, we could add `$(MAKE) -C $(top_srcdir) all` into the rules of `testing` target
> 
> Actually, it is probably simpler to add that as the make rule for the `libyaksa.la` target.

Ahh, we already do that with `libdtpools.la`, certainly we should do the same for `libyaksa.la` :)

hzhou:
test:yaksa

hzhou:
test:yaksa

hzhou:
test:yaksa/whitespace

hzhou:
test:yaksa/whitespace

hzhou:
test:yaksa/whitespace

hzhou:
test:yaksa/whitespace

pavanbalaji:
I see this error with this PR:

```
% ./autogen.sh 

=== generating configure files in main directory ===
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'm4'.
libtoolize: copying file 'm4/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
autoreconf: running: /homes/balaji/software/tools/install/x86_64/bin/autoconf --force
autoreconf: running: /homes/balaji/software/tools/install/x86_64/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:82: installing 'm4/ar-lib'
configure.ac:79: installing 'm4/compile'
configure.ac:84: installing 'm4/config.guess'
configure.ac:84: installing 'm4/config.sub'
configure.ac:75: installing 'm4/install-sh'
configure.ac:75: installing 'm4/missing'
Makefile.am: installing 'm4/depcomp'
autoreconf: Leaving directory `.'
=== done === 

generating backend pup functions for seq... done
generating backend pup functions for cuda... done
generating simple tests ... done
generating pack tests ... Traceback (most recent call last):
  File "./maint/gentests.py", line 140, in <module>
    gen_pack_iov_tests("pack", "test/pack/testlist.gen", "create")
  File "./maint/gentests.py", line 69, in gen_pack_iov_tests
    outfile.write("prefix: " + m[1] + "\n\n")
TypeError: '_sre.SRE_Match' object is not subscriptable
Generating a helper maint/Version... done
```

hzhou:
> generating pack tests ... Traceback (most recent call last):
>   File "./maint/gentests.py", line 140, in <module>
>     gen_pack_iov_tests("pack", "test/pack/testlist.gen", "create")
>   File "./maint/gentests.py", line 69, in gen_pack_iov_tests
>     outfile.write("prefix: " + m[1] + "\n\n")
> TypeError: '_sre.SRE_Match' object is not subscriptable

Which versions of python3 do you use? Direct subscript access for `re` match is only supported after newer version (3.6 I believe). Anyway, I'll updated with older syntax.

pavanbalaji:
For developer tools, such as `gentests.py`, I don't mind having a dependency on very new tools.  But it should fail with a more understandable error.

hzhou:
test:yaksa

hzhou:
test:yaksa

hzhou:
The review trigger is not working because Jenkins report "Exceeded rate limit for repository", which I suspect it's a bogus one.

hzhou:
@pavanbalaji Could you check if it is working for you now?

pavanbalaji:
I think this PR does a number of unrelated changes.  Can you separate the core part that handles the dependencies from the other changes?  We can discuss the other changes separately, which I don't want to mix with the core issue of dependency tracking.

hzhou:
> I think this PR does a number of unrelated changes. Can you separate the core part that handles the dependencies from the other changes? We can discuss the other changes separately, which I don't want to mix with the core issue of dependency tracking.

I kinda anticipated this :) I'll work on it, but it may take some time.

hzhou:
> @hzhou I noticed this PR currently has two problems in loss of functionality:
> 
>     1. I cannot go to a sub-directory (such as `test/pack`) and do `make testing` for that sub-dir.
> 
This should still work. To check, there is static `Makefile` in each test sub-directory with a `testing` make target.

>     2. It lost the capability to go into a sub-dir and make all or individual test binaries. The behavior of `make` is equal to `make testing` now.

We certainly can add "all" and individual test binaries targets into the static Makefile, but that is pure maintenance overhead. I don't think to bring back `Makefile.am` is ideal as that will get a lot of "SUBDIR" noise for a normal build. 

Currently `runtests.py` make individual test binary before each testlist entry automatically. What I think it's a nice feature is to aggregate all test binaries and run `make [listof test binaries]` before running individual tests. Then we can easily add a target to invoke `runtests.py ... --make`, which just to do the make phase.

hzhou:
Replaced by #23

PR #22, opened 2020-04-04
test: update runtests.py

hzhou:
## Pull Request Description

Update `test/runtests.py` with following changes:

* load all tests into tests as list
* store each test results into the test dict
* support specical directives: prefix, condition, testlist, and echo
* support multiple testlist files on the command line

Note: this is a split and prereqisite for PR #10.

## Author Checklist
* [ ] Passes all tests

### DIFF
--- a/maint/gentests.py
+++ b/maint/gentests.py
@@ -1,10 +1,11 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 ##
 ## Copyright (C) by Argonne National Laboratory
 ##     See COPYRIGHT in top-level directory
 ##
 
 import sys
+import re
 
 
 ##### global settings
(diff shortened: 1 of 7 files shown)


hzhou:
test: yaksa

hzhou:
@pavanbalaji This is my attempt of the split. The main goal is to let runtests.py pre-load all tests into a testlist, so it can be flexible in whether we want to do it recusively or integrally. But once we let go of the recursive nature, some of the directives in the testlist become necessary.

pavanbalaji:
@hzhou This is the part that I mentioned was orthogonal to the core change.  Can you do the core fix first?  We can discuss this part later, because it includes many things that are just personal preferences and are different from person to person.

hzhou:
If we let go of the recursive make structure, then we have to load all testlists together, right? For example, we can't simply descend down the directory and do sub-make.

pavanbalaji:
> If we let go of the recursive make structure, then we have to load all testlists together, right? For example, we can't simply descend down the directory and do sub-make.

This PR does more than remove the recursive make structure.  Can you split out that part?

hzhou:
> This PR does more than remove the recursive make structure. Can you split out that part?

This PR still reserves the recursive make structure.

pavanbalaji:
> > This PR does more than remove the recursive make structure. Can you split out that part?
> 
> This PR still reserves the recursive make structure.

This PR does more things that I'm comfortable signing off right now.  That's the reason I'm asking you to separate out things that you consider cleanup, from things that are needed to fix the dependency issue.

PR #33, opened 2020-04-07
Multi-backend-device packing/unpacking

pavanbalaji:
## Pull Request Description

This PR introduces a major new feature that allows packing/unpacking between host and GPU memories.  It provides a full spectrum of pack/unpack operations from any input buffer to any output buffer in the process virtual address space, except for multi-GPU systems.

It separates out "native" operations from "hybrid" operations.  Native operations are straight device-to-device or host-to-host pack/unpack and would use a single backend device.  Hybrid operations, on the other hand, require multiple backend devices.  This patch also introduces a progress engine and a corresponding state machine to handle such multi-device data movement.

This PR never uses `cudaMemcpyAsync` directly from unregistered host buffers.  Data is always copied into temporary registered buffers.  This is because of quirky behavior associated with `cudaMemcpy*` routines when dealing with unregistered buffers (see known shortcomings below for more details).

The following modes are covered for both packing and unpacking --

- Device memory to device memory
- Host memory to host memory
- Device to host registered memory
- Device to host unregistered memory
- Registered host memory to device memory
- Unregistered host memory to device memory

Known shortcomings --

- If the user uses an unregistered memory buffer for the pack or unpack, this PR always performs a copy to a temporary registered buffer.  Depending on the size of the data, the copy itself can be more expensive than registering the user buffer.  We should investigate approaches for dynamically pinning user memory and maintaining a cache of such pinned memory (similar to what MPI implementations do), but this PR does not handle any of that.
- If the user program copies data from unregistered memory buffers to the device using `cudaMemcpy` before calling `yaksa_ipack` or `yaksa_iunpack` for a device-to-device pack/unpack, (s)he needs to be aware that `cudaMemcpy` from unregistered memory  buffers might return before the data is actually copied to the device.  This seems to be a strange, unexpected, behavior, but is what is defined in the CUDA documentation (https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior).  This PR assumes that the user program is correct with the necessary synchronization (e.g., `cudaDeviceSynchronize` is done after the `cudaMemcpy`, or `cudaMemcpy` is performed from a registered memory region), but it is easy for the user to write an incorrect application in this regard.  Alternatively, we could revert PR #32, so yaksa operations are always ordered with respect to the null stream operations (including `cudaMemcpy`).

## Expected Impact

More seamless packing/unpacking irrespective of the location of the memory being used for the source and destination buffers.

## Author Checklist
* [x] Passes all tests


pavanbalaji:
Running jenkins tests for this PR as a sanity check, but this PR really needs to be tested on a GPU system to ensure correctness.  For non-GPU systems, most of the progress engine is disabled anyway, so none of that would get tested.  I'll run the GPU tests manually and report my findings here.

pavanbalaji:
Tagging @gcongiu for his comments on the known shortcomings with respect to PR #32 and `cudaMemcpy`.

pavanbalaji:
test:yaksa/review

pavanbalaji:
test:yaksa

gcongiu:
@pavanbalaji I don't think we should rely on this behavior of CUDA implicitly synchronizing with the default stream. I think it is probably better to always use independent streams for memory transfers between host and device and require the user to do the right thing by calling the proper synchronization routines such as `yaksa_request_wait` in the case of a host-to-device pack, for example, before issuing work to the device for that data. Even for device-to-device transfers yaksa should probably use `cudaMemcpyAsync` and synchronize in place instead of using the default stream.

pavanbalaji:
> @pavanbalaji I don't think we should rely on this behavior of CUDA implicitly synchronizing with the default stream. I think it is probably better to always use independent streams for memory transfers between host and device and require the user to do the right thing by calling the proper synchronization routines such as `yaksa_request_wait` in the case of a host-to-device pack, for example, before issuing work to the device for that data. Even for device-to-device transfers yaksa should probably use `cudaMemcpyAsync` and synchronize in place instead of using the default stream.

@gcongiu You are misunderstanding the issue here.  Consider this program:

```
sbuf = malloc(size);
cudaMalloc(tbuf, size);
cudaMalloc(dbuf, size);

initialize_buffer(sbuf);
cudaMemcpy(tbuf, sbuf, size, cudaMemcpyHostToDevice);
yaksa_ipack(tbuf, dbuf, ...);
yaksa_wait(...);
```

This program as it stands is incorrect because of the quirky semantics of `cudaMemcpy`.

gcongiu:
@pavanbalaji I understand that (I should have probably clarified better). I am saying the user should not rely on Yaksa using CUDA implicit synchronization (by assuming it creates internal streams using `cudaStreamCreate`) with stream 0 to make that code work as this is a CUDA specifics and the code is incorrect anyway. The user should do the right thing, for example, using `cudaDeviceSynchronize`, as you mentioned in the description of the PR. My impression was that you were suggesting the option to reintroduce implicit synchronization to make user's mistakes work as well. This will probably not work anyway if, for example, the program is multithreaded and operations are not ordered among the threads (I know this is probably not the common use case but it might happen?).

src/backend/src/yaksuri_progress.c
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    if (nelems > elem->pup.count - elem->pup.completed_count - elem->pup.issued_count)
+        nelems = elem->pup.count - elem->pup.completed_count - elem->pup.issued_count;
+
+    /* if we don't have enough space, return */
+    if (nelems == 0) {
+        goto fn_exit;
+    }
+
```
yfguo, line 203: Please remove double empty lines.

src/backend/src/yaksuri_progress.c
```diff
@@ -0,0 +1,532 @@
...
+    if (need_host_buf) {
+        if (yaksuri_global.cuda.host.slab == NULL) {
+            yaksuri_global.cuda.host.slab = yaksuri_global.cuda.host.malloc(TMPBUF_SLAB_SIZE);
+        }
+        yaksuri_global.cuda.host.slab_tail_offset =
+            host_tmpbuf_offset + nelems * elem->pup.type->size;
+    }
+
```
yfguo, line 221: Please remove double empty lines.

src/backend/src/yaksuri_progress.c
```diff
@@ -0,0 +1,532 @@
...
+            progress_subop_s *tmp = subop->next;
+            rc = free_subop(subop);
+            YAKSU_ERR_CHECK(rc, fn_fail);
+            subop = tmp;
+        } else {
+            break;
+        }
+    }
+
```
yfguo, line 379: Please remove double empty lines.

src/backend/src/yaksuri_progress.c
```diff
@@ -0,0 +1,532 @@
...
+    /****************************************************************************/
+    /* Step 2: If we don't have any more work to do, return */
+    /****************************************************************************/
+    if (elem->pup.completed_count == elem->pup.count) {
+        rc = progress_dequeue(elem);
+        YAKSU_ERR_CHECK(rc, fn_fail);
+        goto fn_exit;
+    }
+
```
yfguo, line 389: Please remove double empty lines.

src/backend/src/yaksuri_progress.c
pavanbalaji: The double empty lines are intentional to separate logically different parts of the code.

pavanbalaji:
> Running jenkins tests for this PR as a sanity check, but this PR really needs to be tested on a GPU system to ensure correctness. For non-GPU systems, most of the progress engine is disabled anyway, so none of that would get tested. I'll run the GPU tests manually and report my findings here.

Testing on a GPU cluster did not show any errors.

PR #36, opened 2020-04-09
test: free description string

gcongiu:
## Pull Request Description

dtpools no longer manages the description string. This is now generated
on demand by tests for single cases in which debugging info is needed.
The description string is thus handled by the test and should be freed
afterwards.

## Author Checklist

### DIFF
--- a/test/flatten/flatten.c
+++ b/test/flatten/flatten.c
@@ -72,8 +72,6 @@ int main(int argc, char **argv)
     assert(rc == DTP_SUCCESS);
 
     for (int i = 0; i < iters; i++) {
-        char *desc;
-
         dprintf("==== iter %d ====\n", i);
 
         /* create the source object */
@@ -84,9 +82,11 @@ int main(int argc, char **argv)
         assert(sbuf);
 
         if (verbose) {
+            char *desc;
             rc = DTP_obj_get_description(sobj, &desc);
             assert(rc == DTP_SUCCESS);
             dprintf("==> sbuf %p, sobj (count: %zu):\n%s\n", sbuf, sobj.DTP_type_count, desc);
+            free(desc);
         }
 
         rc = DTP_obj_buf_init(sobj, sbuf, 0, 1, basecount);
--- a/test/iov/iov.c
+++ b/test/iov/iov.c
@@ -138,8 +138,6 @@ int main(int argc, char **argv)
     uintptr_t *segment_lengths = (uintptr_t *) malloc(segments * sizeof(uintptr_t));
 
     for (int i = 0; i < iters; i++) {
-        char *desc;
-
         dprintf("==== iter %d ====\n", i);
 
         /* create the source object */
@@ -150,10 +148,12 @@ int main(int argc, char **argv)
         assert(sbuf);
 
         if (verbose) {
+            char *desc;
             rc = DTP_obj_get_description(sobj, &desc);
             assert(rc == DTP_SUCCESS);
             dprintf("==> sbuf %p (with offset) %p, sobj (count: %zu):\n%s\n",
                     sbuf, sbuf + sobj.DTP_buf_offset, sobj.DTP_type_count, desc);
+            free(desc);
         }
 
         rc = DTP_obj_buf_init(sobj, sbuf, 0, 1, basecount);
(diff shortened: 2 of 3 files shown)


gcongiu:
test:yaksa

gcongiu:
@pavanbalaji I missed this in mpich as well. Will fix that too. Thanks @hzhou for pointing this out.

hzhou:
@gcongiu Why don't you keep the description pointer with the dtp_obj, let ` DTP_obj_get_description(obj)` directly return a `char *`, and free the description point inside `DTP_obj_free`?

pavanbalaji:
The branch seems to be out of date.  Can you rebase it on to `master` and rerun the tests?

pavanbalaji:
> @gcongiu Why don't you keep the description pointer with the dtp_obj, let ` DTP_obj_get_description(obj)` directly return a `char *`, and free the description string inside `DTP_obj_free`? Internally you can easily tell whether the description string has been allocated or not.

I think @gcongiu's approach is cleaner.

hzhou:
> > @gcongiu Why don't you keep the description pointer with the dtp_obj, let ` DTP_obj_get_description(obj)` directly return a `char *`, and free the description string inside `DTP_obj_free`? Internally you can easily tell whether the description string has been allocated or not.
> 
> I think @gcongiu's approach is cleaner.

IMO, having the  memory allocated and freed across a library and application is less clean,  but anyway :)

pavanbalaji:
> IMO, having the memory allocated and freed across a library and application is less clean, but anyway :)

Yeah, that's fair.  Perhaps it's cleaner to change `get_description` to `desc_alloc` and add a new function for `desc_free` to be consistent?

hzhou:
> > IMO, having the memory allocated and freed across a library and application is less clean, but anyway :)
> 
> Yeah, that's fair. Perhaps it's cleaner to change `get_description` to `desc_alloc` and add a new function for `desc_free` to be consistent?

The way I see it, application never want to take the ownership of the description string, so if dtpool just keep the pointer and lazily allocate it and free it on clean-up, the interface is much simpler.

pavanbalaji:
> The way I see it, application never want to take the ownership of the description string, so if dtpool just keep the pointer and lazily allocate it and free it on clean-up, the interface is much simpler.

No, it is taking ownership of the description string.  It is seeing it and using it.  This function is not different from other memory allocation functions, except that it returns an initialized buffer.

pavanbalaji:
One alternative would be to completely hide the pointer and allow it to be accessed only through functions.  Then the user won't have access to the pointer at all, and dtpools can figure out when to allocate or free it.

hzhou:
> One alternative would be to completely hide the pointer and allow it to be accessed only through functions. Then the user won't have access to the pointer at all, and dtpools can figure out when to allocate or free it.

I thought that was what I was proposing -- user access it through `DTP_obj_get_description(obj)`.

pavanbalaji:
> > One alternative would be to completely hide the pointer and allow it to be accessed only through functions. Then the user won't have access to the pointer at all, and dtpools can figure out when to allocate or free it.
> 
> I thought that was what I was proposing -- user access it through `DTP_obj_get_description(obj)`.

You said `let DTP_obj_get_description(obj) directly return a char *`.  That's not what I was proposing.  I was proposing not exposing the pointer at all.  So, the function would change to `DTP_obj_print_description`.  There's no pointer exposed.  It has many shortcomings, of course, including what file pointer the data should go into.

Once the user sees `char *`, the ownership becomes less clear.  It becomes closer to `calloc`, which basically says "here's an initialized pointer, which you are responsible for cleaning up".

hzhou:
What about return a `const char *` ? 😄

pavanbalaji:
> What about return a `const char *` ? 😄

That's not what `const` says in the C specification.

hzhou:
I'll let you decide. 😄

pavanbalaji:
> I'll let you decide. 😄

If I had good ideas for this, I'd have done them already.  😃 

Your comments are well taken.  I just don't have a good fix for them.  Making everything explicit (similar to `calloc`) seems to be somewhat clean.

gcongiu:
@hzhou that is probably a better option

gcongiu:
Alright, I will leave it like this

gcongiu:
test:yaksa

Issue #19, opened 2020-04-02
CUDA: automatic SM detection


pavanbalaji:
The current yaksa build system requires the user to provide the actual SM required to compile.  This is cumbersome and puts too much burden on the end user.  It'd be cleaner if the build system could automatically detect the SM and use it.

One option would be to use a shotgun approach and build a fat binary with all possible SMs, but that increases the build time significantly.

gcongiu:
What about detecting the attached devices using `lspci` and matching these against some available compute capability map during configure? One problem with this approach is that users will have to compile the code on the compute nodes they wish to run in order for it to work. Maybe the best option is to request the user to run a script that extracts the needed information on the compute nodes and pass the output of such script to the configure?

pavanbalaji:
> What about detecting the attached devices using `lspci` and matching these against some available compute capability map during configure? One problem with this approach is that users will have to compile the code on the compute nodes they wish to run in order for it to work.

Right, that's a significant limitation.  I wouldn't prefer this approach.

> Maybe the best option is to request the user to run a script that extracts the needed information on the compute nodes and pass the output of such script to the configure?

Providing a script to tell the user what SM to use is useful to have, but I was hoping that we could do better than that.  We might need to contact the NVIDIA folks to get an answer on this.

As a side note, our configure can probably be improved somewhat by allowing for a more human-readable format, such as GPU type (e.g., p100).

pavanbalaji:
One possibility to consider is to use something like `nvrtc` to compile the kernel at datatype creation time.  That way, we don't need to know the SM information till runtime.  A side benefit of this approach would be that we won't need to autogenerate all the cases, and instead generate targeted kernels for the exact datatypes that the user creates.

pavanbalaji:
As far as I can tell, folks seem to be recommending building fat binaries for this (e.g., https://forums.developer.nvidia.com/t/auto-detect-gpu-arch-in-make-file/75289).  I guess we could use that as the default, with an option for the user/developer to override it at configure time.

PR #40, opened 2020-04-12
CUDA fat binaries

pavanbalaji:
## Pull Request Description

Allow the CUDA backend to build executables for multiple architectures, including all supported architectures depending on the CUDA version.  Users still have the option to specialize builds for specific architectures, but are not required to do so.

## Expected Impact

Adds convenience for the user so they do not have to figure out the exact GPU architecture they are building for.  Also helps binary distributors where the binary code needs to work on multiple architectures.

On the negative side, this significantly increases the build time.  In my experiments, the default (parallel) build goes up from 35 seconds to 315 seconds.

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -3,13 +3,57 @@
 ##     See COPYRIGHT in top-level directory
 ##
 
+
+##########################################################################
+##### capture user arguments
+##########################################################################
+
 # --with-cuda-sm
-AC_ARG_WITH([cuda-sm],AS_HELP_STRING([--with-cuda-sm=<numeric>],[builds CUDA support for that architecture]),,
-            [with_cuda_sm=none])
-if test "${with_cuda_sm}" != "none" ; then
-    CUDA_SM="${with_cuda_sm}"
-    AC_SUBST(CUDA_SM)
-fi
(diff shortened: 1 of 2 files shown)


pavanbalaji:
This PR fixes pmodels/yaksa#19.

pavanbalaji:
@gcongiu Can you ignore the first two patches in this PR?  They are from the PR pmodels/yaksa#39, which is being reviewed separately.

src/backend/cuda/subconfigure.m4
```diff
@@ -41,3 +77,121 @@ elif test "${enable_cuda_p2p}" = "cliques" ; then
...
+                    PAC_APPEND_FLAG([53],[CUDA_SM])
+                fi
+
+                if test ${cuda_version} -ge 6 ; then
+                    # kepler
+                    PAC_APPEND_FLAG([30],[CUDA_SM])
+                    PAC_APPEND_FLAG([35],[CUDA_SM])
+                    PAC_APPEND_FLAG([37],[CUDA_SM])
+                fi
+                ;;
```
gcongiu, line 137: Nit pick: one branch

pavanbalaji: Yikes, it should be `-ge 5` for `kepler`.  Thanks for catching it.  I'll fix it.

Issue #17, opened 2020-04-02
CUDA: respect maximum number of thread blocks


pavanbalaji:
The current Yaksa code does not respect the maximum number of thread blocks allowed in CUDA.

Since compute capability 3.0, the maximum number of thread blocks in the X dimension can go up to a maximum of `2^31 - 1` (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications).  Yaksa could potentially use the Y and Z dimensions of the thread blocks to increase the maximum number of thread blocks it can use, but those are limited to 64K each as well.

In the current code, yaksa uses only the X dimension of thread blocks.  With around 256 threads per thread block, and each thread packing or unpacking one element of data, we can pack/unpack up to 512 billion basic datatype elements (at least 512GB, if the basic datatype is a byte).  Given that the user would need to supply a source buffer and a destination buffer, this can only be a problem on systems that have more than 1TB of free memory for the application to use.

This might be sufficient for now, but it is somewhat close to what current systems already provide, especially when we take persistent memory into account.

It would be better if we can use the Y and Z dimensions too.  That'd increase the maximum data that can be packed/unpacked to 2 Yottabytes (times the basic datatype size).

Issue #30, opened 2020-04-06
CUDA backend performance tuning


yfguo:
We need to investigate and study the best strategy for performance tuning in the CUDA backend.

One knob is the thread block size vs number of blocks.

PR #41, opened 2020-04-13
CUDA: Allow for more thread blocks than the X dimension of the block grid

pavanbalaji:
## Pull Request Description

This PR allows us to have as many thread blocks as allowed in all of the three dimensions of the block grid combined.

## Expected Impact

This would allow us to pack/unpack larger data sizes than before.

## Author Checklist

### DIFF
--- a/src/backend/cuda/include/yaksuri_cudai.h
+++ b/src/backend/cuda/include/yaksuri_cudai.h
@@ -85,9 +85,9 @@ typedef struct yaksuri_cudai_md_s {
 
 typedef struct yaksuri_cudai_type_s {
     void (*pack) (const void *inbuf, void *outbuf, uintptr_t count, yaksuri_cudai_md_s * md,
-                  int n_threads, int n_blocks, int device);
+                  int n_threads, int n_blocks_x, int n_blocks_y, int n_blocks_z, int device);
     void (*unpack) (const void *inbuf, void *outbuf, uintptr_t count, yaksuri_cudai_md_s * md,
-                    int n_threads, int n_blocks, int device);
+                    int n_threads, int n_blocks_x, int n_blocks_y, int n_blocks_z, int device);
     yaksuri_cudai_md_s *md;
     pthread_mutex_t mdmutex;
     uintptr_t num_elements;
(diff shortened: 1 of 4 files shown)


pavanbalaji:
This PR fixes pmodels/yaksa#17

src/backend/cuda/include/yaksuri_cudai.h
```diff
@@ -16,6 +16,9 @@
 #define CUDA_P2P_CLIQUES  (3)
 
 #define YAKSURI_CUDAI_THREAD_BLOCK_SIZE  (256)
```
gcongiu, line 18: The max x or y block size is 1024. Is this lower than that because you want to allow the pack/unpack to share SMs with user kernels? Smaller block size will imply more context switches among the thread blocks, thus, if sharing is not a concern maybe this should be increased?

pavanbalaji: Right.  There's a separate issue for tuning the number of threads and block sizes (pmodels/yaksa#30).

gcongiu: Thanks! I missed that

src/backend/cuda/pup/yaksuri_cudai_pup.c
```diff
@@ -9,6 +9,42 @@
...
+static int get_thread_block_dims(uint64_t count, yaksi_type_s * type, int *n_threads,
+                                 int *n_blocks_x, int *n_blocks_y, int *n_blocks_z)
+{
+    int rc = YAKSA_SUCCESS;
+    yaksuri_cudai_type_s *cuda_type = (yaksuri_cudai_type_s *) type->backend.cuda.priv;
+
+    *n_threads = THREAD_BLOCK_SIZE;
+    uint64_t n_blocks = count * cuda_type->num_elements / THREAD_BLOCK_SIZE;
+    n_blocks += ! !(count * cuda_type->num_elements % THREAD_BLOCK_SIZE);
+
```
gcongiu, line 26: For correctness, should this return an error code if the number of blocks exceeds the max allowed size? Or simply assert?

gcongiu: Oh I think I have commented too late

pavanbalaji: That would be more than the size of `int64_t`.  At that point, we'd need to change a whole lot of code in yaksa to make it work, and an assert would not be sufficient.

PR #43, opened 2020-04-13
Allow users to control the maximum number of nesting levels generated for the pup functions

pavanbalaji:
## Pull Request Description

Allow for users to generate different levels of nesting, thus trading off performance with the library size and build time.  The default is set to 3 levels of nesting, as before.

## Author Checklist

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -4,6 +4,10 @@
 ##     See COPYRIGHT in top-level directory
 ##
 
+########################################################################
+## Utility functions
+########################################################################
+
 echo_n() {
     # "echo -n" isn't portable, must portably implement with printf
     printf "%s" "$*"
@@ -14,6 +18,10 @@ error() {
 }
 
 
+########################################################################
+## Autotools
+########################################################################
+
 # generate configure files
 echo
 echo "=== generating configure files in main directory ==="
@@ -21,6 +29,10 @@ autoreconf -vif
 echo "=== done === "
 echo
 
+
+########################################################################
+## Generating required files
+########################################################################
 
 # backend pup functions
 for x in seq cuda ; do
--- a/autogen.sh
+++ b/autogen.sh
@@ -15,7 +15,26 @@ echo_n() {
 
 error() {
     echo "===> ERROR:   $@"
+    exit
 }
+
+
+########################################################################
+## Parse user arguments
+########################################################################
+
+genpup_args=
+for arg in "$@" ; do
+    case $arg in
+        -pup-max-nesting=*|--pup-max-nesting=*)
+            genpup_args="$genpup_args $arg"
+            ;;
+
+        *)
+            error "unknown argument $arg"
+            ;;
+    esac
+done
 
 
 ########################################################################
@@ -37,7 +56,7 @@ echo
 # backend pup functions
 for x in seq cuda ; do
     echo_n "generating backend pup functions for ${x}... "
-    ./src/backend/${x}/genpup.py
+    ./src/backend/${x}/genpup.py ${genpup_args}
     echo "done"
 done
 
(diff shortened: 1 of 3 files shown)


autogen.sh
```diff
@@ -14,6 +18,24 @@ error() {
...
+## Parse user arguments
+########################################################################
+
+pup_max_nesting=3
+while test ! -z "$1" ; do
+    if test "$1" = "--pup-max-nesting" ; then
+        shift
+        pup_max_nesting=$1
+	shift
+    fi
```
hzhou, line 31: It's a dead loop if wrong argument is supplied since `$1` will never get shifted.

autogen.sh
```diff
@@ -22,10 +44,14 @@ echo "=== done === "
...
 
+########################################################################
+## Generating required files
+########################################################################
+
 # backend pup functions
 for x in seq cuda ; do
     echo_n "generating backend pup functions for ${x}... "
-    ./src/backend/${x}/genpup.py
+    ./src/backend/${x}/genpup.py --pup-max-nesting=${pup_max_nesting}
```
hzhou, line 54: Would it be simpler if you use the same option syntax for `autogen.sh` and `genpup.py`? You can use a case switch and collect the arguments as is and pass them to `genpup.py`.

src/backend/cuda/genpup.py
```diff
@@ -320,7 +321,16 @@ def switcher(typelist, pupstr, nests):
 ##### main function
 ########################################################################################
 if __name__ == '__main__':
-    #### generate the core pack/unpack kernels
+    ##### parse user arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--pup-max-nesting', type=int, help='maximum nesting levels to generate', required=True)
+    args = parser.parse_args()
```
hzhou, line 327: How about add some sanity check so the nesting level is not negative or too big? I think it is nice to have at least a warning when the nesting level is above e.g. 4 so make sure the user is aware of what is going to happen.

autogen.sh
pavanbalaji: Good catch.  I'll fix it.

autogen.sh
pavanbalaji: For now, the arguments are only for `genpup.py`, but `autogen.sh` runs more than that one script (for example, `gentests.py`).  So we can't simply pass all arguments to `genpup.py`.

src/backend/cuda/genpup.py
pavanbalaji: I can check for negative values.  But "too large" is an arbitrary number, which I don't want to add.

autogen.sh
hzhou: I understand. I mean something like this:
```
genpup_args=
for a in $@ ; do
    case a in 
        --pup-*) genpup_args="$genpup_args $a" ;;
    esac
done
```
then
```
    eval ./src/.../genpup.py $genpup_args
```

pavanbalaji: Yeah, that's a good idea.  It'll also naturally separate the namespace for the arguments for different scripts.  I'll fix it.

src/backend/seq/genpup.py
```diff
@@ -340,7 +341,21 @@ def switcher(typelist, pupstr, nests):
...
+        parser.print_help()
+        print
+        print("===> ERROR: pup-max-nesting must be positive")
+        sys.exit(1)
+
+    ##### generate the list of derived datatype arrays
+    darraylist = [ ]
+    yutils.generate_darrays(derived_types, darraylist, args.pup_max_nesting)
+
+    ##### generate the core pack/unpack kernels
```
hzhou, line 358: Is it easy to factor out a common module and avoid duplication?

pavanbalaji: `yutils` is the common module.  It's meant for functions that are generic across all `yaksa` scripts.

Issue #21, opened 2020-04-03
backend: optimize yaksuri_seqi_populate_pupfns for compilation


hzhou:
Currently, we generate this humongous `yaksuri_seqi_populate_pupfns` function, which is 159800 lines of nested switch cases and compilers are having a very hard time optimizing it. This single file takes many minutes just to compile and become the bottleneck for the development cycle.

Can we do something to speed up the compilation without degrading the performance?

Floating ideas around:
* split it into several smaller functions -- so the compiler stop attempting to juggling all parts together
* calculate hash from yaksi_type_s and use the hash to find the entry

hzhou:
`time make src/backend_seq/pup/yasuri_seqi_pub.lo`:

* -O2: 8 min 25 sec
* -O1: 2 min  7 sec
* -O0: 9 sec

yfguo:
I was building the latest code with `./configure --prefix=$PWD/_inst --with-cuda=/opt/cuda --with-cuda-sm=61` on a 32GB memory machine. The build run out of memory at `CC src/backend/seq/pup/yaksuri_seqi_populate_pupfns.lo`.

PR #45, opened 2020-04-14
Populate pupfns improvements for build time

pavanbalaji:
## Pull Request Description

This PR splits the populate pupfns routine into multiple smaller routines to help with the build time.

## Expected Impact

Significantly faster build time.  With the Intel compiler, the (parallel) build time is down from 21 minutes to around 1 minute with this PR.

Fixes pmodels/yaksa#21

## Author Checklist


src/backend/gencomm.py
```diff
@@ -0,0 +1,252 @@
...
+    yutils.display(OUTFILE, "}\n")
+
+
+########################################################################################
+##### main function
+########################################################################################
+def populate_pupfns(pup_max_nesting, backend, blklens, builtin_types, builtin_maps):
+    ##### generate the list of derived datatype arrays
+    darraylist = [ ]
+    yutils.generate_darrays(derived_types, darraylist, pup_max_nesting)
```
hzhou, line 101: `darraylist` isn't needed here, right?

src/backend/seq/genpup.py
```diff
@@ -6,17 +6,22 @@
...
+########################################################################################
+##### Global variables
+########################################################################################
 
-def display(*argv):
-    for x in range(indentation):
-        OUTFILE.write("    ")
-    for arg in argv:
-        OUTFILE.write(arg)
+num_paren_open = 0
```
hzhou, line 21: You already have `yutils.display` to track indentation based on braces, it should be easy enough to let it also track `num_parent_open`, and say, let `yutils.display(OUTFILE, "WRAP_BRACES")` to automatically wrap all open braces. In fact, aren't `num_paren_open` always the same as `display.indent`?

src/backend/seq/genpup.py
```diff
@@ -138,6 +131,7 @@ def resized(suffix, b, blklen, last):
     "contig": contig,
     "resized": resized,
 }
+
```
hzhou, line 134: Can we move `derived_maps` and maybe `builtin_maps` into `gencomm.py`?

And for `derived_decl_maps`, you could use [getattr](https://docs.python.org/library/functions.html#getattr) to get the function from a string instead of hardcode the map. Example: https://stackoverflow.com/questions/3061/calling-a-function-of-a-module-by-using-its-name-a-string

.gitignore
```diff
@@ -86,3 +56,123 @@ missing
...
+/src/backend/seq/pup/yaksuri_seqi_pup_int.c
+/src/backend/seq/pup/yaksuri_seqi_pup_int16_t.c
+/src/backend/seq/pup/yaksuri_seqi_pup_int32_t.c
+/src/backend/seq/pup/yaksuri_seqi_pup_int64_t.c
+/src/backend/seq/pup/yaksuri_seqi_pup_int8_t.c
+/src/backend/seq/pup/yaksuri_seqi_pup_long.c
+/src/backend/seq/pup/yaksuri_seqi_pup_long_double.c
+/src/backend/seq/pup/yaksuri_seqi_pup_long_long.c
+/src/backend/seq/pup/yaksuri_seqi_pup_short.c
+/src/backend/seq/pup/yaksuri_seqi_pup_wchar_t.c
```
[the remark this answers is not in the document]
pavanbalaji: Too risky, because not all files are autogenerated.

src/backend/gencomm.py
```diff
@@ -0,0 +1,252 @@
+#! /usr/bin/env python
```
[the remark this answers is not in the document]
pavanbalaji: That's a bad idea.  Scripts at the top-level should *NOT* know about internals.  They should be placed at the level they have the knowledge for.

autogen.sh
```diff
@@ -64,6 +52,18 @@ done
...
+########################################################################
+
+# generate configure files
+echo
+echo "=== generating configure files in main directory ==="
+autoreconf -vif
+echo "=== done === "
+echo
+
+
```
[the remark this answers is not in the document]
pavanbalaji: yup, will do.

maint/yutils.py
```diff
@@ -6,6 +6,22 @@
...
+            display.indent -= 1
+    for x in range(display.indent):
+        OUTFILE.write("    ")
+    for arg in argv:
+        OUTFILE.write(arg)
+        if (arg.find('{') != -1):
+            display.indent += 1
+display.indent = 0
+
+
```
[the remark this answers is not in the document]
pavanbalaji: yup, will do.

src/backend/gencomm.py
pavanbalaji: Oh, good catch.  Might have been leftover from a previous version.

src/backend/seq/genpup.py
pavanbalaji: In the current version, yes.  But I don't know how many parenthesis to close in some places.  I think we can leave it in for now.

src/backend/seq/genpup.py
pavanbalaji: `builtin_maps` are different for seq and cuda.  I'll look into `getattr`.

pavanbalaji:
@hzhou Thanks for the comments.  I've split out the commits some more.  Hopefully this version will be easier to review.

maint/yutils.py
```diff
@@ -25,6 +25,7 @@ def copyright(outfile):
 ##### generate an array of datatype arrays
 ########################################################################################
 def generate_darrays(derived_types, darraylist, maxlevels):
+    darraylist.append([])
```
hzhou, line 55: Why do we need this empty entry?

pavanbalaji:
Hold off on this review.  It looks like some unrelated changes crept into the final patch.  I'll clean it up.

maint/yutils.py
pavanbalaji: Earlier, all levels were being generated dynamically.  So we don't have a use case for all levels being "NULL" (basically the builtin case), because that was handled separately as a direct memcpy.

Now we have separated out two of the nesting levels to be generated statically and generate the remaining levels dynamically.  So the dynamically generated levels can have both "NULL" (meaning that only the top two levels are considered for the pack/unpack kernels).

pavanbalaji:
@hzhou  Done cleaning up.  Can you review it now?

PR #58, opened 2020-04-29
Better detect zero-byte types

pavanbalaji:
Look at the count, the blocklengths and the sizes of each of the child
types to decide whether a new type is a zero-byte type.

## Author Checklist


pavanbalaji:
> What if the user passes `YAKSA_TYPE__NULL` to a type create function?

That'd be an error.

PR #60, opened 2020-04-29
Coverity fixes

pavanbalaji:
Fix some of the issues reported by coverity.

## Author Checklist

### DIFF
--- a/src/frontend/types/yaksa_blkindx.c
+++ b/src/frontend/types/yaksa_blkindx.c
@@ -109,6 +109,7 @@ int yaksa_create_hindexed_block(int count, int blocklength, const intptr_t * arr
 int yaksa_create_indexed_block(int count, int blocklength, const int *array_of_displs,
                                yaksa_type_t oldtype, yaksa_type_t * newtype)
 {
+    intptr_t *real_array_of_displs = NULL;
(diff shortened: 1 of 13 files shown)


src/frontend/types/yaksa_blkindx.c
```diff
@@ -131,11 +133,11 @@ int yaksa_create_indexed_block(int count, int blocklength, const int *array_of_d
     rc = yaksi_create_hindexed_block(count, blocklength, real_array_of_displs, intype, &outtype);
     YAKSU_ERR_CHECK(rc, fn_fail);
 
-    free(real_array_of_displs);
-
     *newtype = outtype->id;
 
   fn_exit:
+    if (real_array_of_displs)
```
gcongiu, line 139: do you need this check?

pavanbalaji: Yeah, in the error case, we might end up here before the array is actually allocated.

gcongiu: `free(NULL)` is safe right?

pavanbalaji: No, but we check for that case when freeing the type.

PR #70, opened 2020-05-05
build: ignore unrecognized options to --enable-fast

pavanbalaji:
## Pull Request Description

In embedded builds, the user might provide extra options to `--enable-fast`.  Ignore them.

## Author Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -117,10 +117,10 @@ fi
 
 # --enable-fast
 AC_ARG_ENABLE([fast],AS_HELP_STRING([--enable-fast=O<opt>],[adds -O<opt> to CFLAGS]),,[enable_fast=yes])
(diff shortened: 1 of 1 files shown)


pavanbalaji:
> I see. I think the patch is fine. But maybe a better fix is at `mpich` to filter some of the options away. MPICH will set the CFLAGS so the submodule will inherit the CFLAGS anyway. And in general, we can't require third-party modules always coordinate with mpich on what flags to honor and what to ignore; probably should be other way around.

Yeah, I'd agree with that.  Could you file an issue for it, so we don't forget about it?  I'll merge this in so we can move forward with the integration for now.

hzhou:
Issue https://github.com/pmodels/mpich/issues/4497

PR #72, opened 2020-05-06
Optimizations for flattening of datatypes

pavanbalaji:
## Pull Request Description

Flattening/unflattening of datatypes can be expensive when there are many child datatypes involved.  This is particularly true for structs, where the number of child datatypes can be arbitrarily large.  This PR optimizes flattening/unflattening in some cases.

## Author Checklist

### DIFF
--- a/src/frontend/flatten/yaksa_unflatten.c
+++ b/src/frontend/flatten/yaksa_unflatten.c
@@ -9,38 +9,36 @@
 #include <string.h>
 #include <stdlib.h>
 
-static int unflatten(yaksi_type_s ** type, const void *flattened_type)
+static inline int unflatten(yaksi_type_s ** type, const void *flattened_type)
 {
     int rc = YAKSA_SUCCESS;
     yaksi_type_s *newtype;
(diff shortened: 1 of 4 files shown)


src/frontend/flatten/yaksa_unflatten.c
```diff
@@ -9,38 +9,35 @@
...
+        YAKSU_ERR_CHECK(rc, fn_fail);
+        yaksu_atomic_incr(&tmp->refcount);
+        newtype = tmp;
+        goto setup_hooks;
+    } else {
+        rc = yaksi_type_alloc(&newtype);
+        YAKSU_ERR_CHECK(rc, fn_fail);
+
+        /* don't overwrite the local ID, but keep the original ID that
+         * we got over the flattened type */
```
gcongiu, line 31: Is this comment for the builtin case?

pavanbalaji: Part of it.  The "don't overwrite the local ID" part should be there.  I'll redo the comment when I repush.  Is that the only comment?

PR #91, opened 2020-06-04
Fix memory leaks in init and pool allocator

raffenet:
## Pull Request Description

Fix remaining leaks found with AddressSanitizer. One in initialization, and one in the pool allocator.

## Expected Impact

Fix memory leaks.

## Author Checklist

### DIFF
--- a/src/util/yaksu_pool.c
+++ b/src/util/yaksu_pool.c
@@ -77,6 +77,7 @@ int yaksu_pool_free(yaksu_pool_s pool)
                 count++;
 
         pool_head->free_fn(tmp->slab);
+        free(tmp->elems);
         free(tmp);
         tmp = next;
     }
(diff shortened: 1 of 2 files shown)


raffenet:
I've got one more commit the fixes the remaining leaks. Will push and update description shortly.

src/util/yaksu_pool.c
```diff
@@ -77,6 +77,7 @@ int yaksu_pool_free(yaksu_pool_s pool)
                 count++;
 
         pool_head->free_fn(tmp->slab);
+        pool_head->free_fn(tmp->elems);
```
pavanbalaji, line 80:
```suggestion
        free(tmp->elems);
```

pavanbalaji: I need to go back and see why my valgrind testing didn't catch any of these errors.  These are fairly obvious errors that valgrind should be able to catch.

raffenet: AddressSanitizer is a wonder. I think `libfabric` enables it be default for debug configurations. Probably a good idea for new projects.

pavanbalaji: I'll add it to yaksa's testing too.

PR #104, opened 2020-08-17
Fixes for 32-bit systems

pavanbalaji:
## Pull Request Description

We calculate the potential extent of a datatype, make sure it fits inside yaksa's extent type format, and then decide whether we can use this datatype or not.  For this, we need the internal calculation of the extent to always be done with 64 bits, so we can eventually check if it is within the required range or not.

## Author Checklist


pavanbalaji:
Restarting the platform tests as `mpich-mac1` seems to have crashed.

test:yaksa/review/platforms

pavanbalaji:
test:yaksa/review/platforms

PR #115, opened 2020-09-12
Lockfree conversion of handle to internal object

pavanbalaji:
## Pull Request Description

Generalize handle management, and create a handle cache, for faster conversion of handles to internal objects.

## Author Checklist

### DIFF
--- a/src/frontend/include/yaksi.h
+++ b/src/frontend/include/yaksi.h
@@ -70,10 +70,6 @@ struct yaksi_request_s;
 typedef struct {
     yaksu_handle_pool_s type_handle_pool;
     yaksu_handle_pool_s request_handle_pool;
-
-    /* keep a cache of the builtin types, so we do not have to get a
-     * lock to query for them each time */
-    struct yaksi_type_s *yaksi_builtin_types[YAKSI_TYPE__LAST];
 } yaksi_global_s;
 extern yaksi_global_s yaksi_global;
 
--- a/src/frontend/include/yaksi.h
+++ b/src/frontend/include/yaksi.h
@@ -42,6 +42,30 @@ typedef enum {
 struct yaksi_type_s;
 struct yaksi_request_s;
 
+/*
+ * The type handle is divided into the following parts:
+ *
+ *    32 bits -- unused
+ *    32 bits -- type object id
+ */
+#define YAKSI_TYPE_UNUSED_BITS                  (32)
+#define YAKSI_TYPE_OBJECT_ID_BITS               (32)
+
+#define YAKSI_TYPE_GET_OBJECT_ID(handle) \
+    ((handle) << (64 - YAKSI_TYPE_OBJECT_ID_BITS) >> (64 - YAKSI_TYPE_OBJECT_ID_BITS))
+
+/*
+ * The request handle is divided into the following parts:
+ *
+ *    32 bits -- unused
+ *    32 bits -- request object id
+ */
+#define YAKSI_REQUEST_UNUSED_BITS                  (32)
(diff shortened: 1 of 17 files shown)


src/frontend/include/yaksi.h
```diff
@@ -42,6 +42,30 @@ typedef enum {
...
+ * The type handle is divided into the following parts:
+ *
+ *    32 bits -- unused
+ *    32 bits -- type object id
+ */
+#define YAKSI_TYPE_UNUSED_BITS                  (32)
+#define YAKSI_TYPE_OBJECT_ID_BITS               (32)
+
+#define YAKSI_TYPE_GET_OBJECT_ID(handle) \
+    (handle << (64 - YAKSI_TYPE_OBJECT_ID_BITS) >> (64 - YAKSI_TYPE_OBJECT_ID_BITS))
```
shintaro-iwasaki, line 55: I recommend `(handle) << ...` rather than `handle <<` to avoid an operator precedence bug.

src/frontend/include/yaksi.h
```diff
@@ -42,6 +42,30 @@ typedef enum {
...
+ * The request handle is divided into the following parts:
+ *
+ *    32 bits -- unused
+ *    32 bits -- request object id
+ */
+#define YAKSI_REQUEST_UNUSED_BITS                  (32)
+#define YAKSI_REQUEST_OBJECT_ID_BITS               (32)
+
+#define YAKSI_REQUEST_GET_OBJECT_ID(handle) \
+    (handle << (64 - YAKSI_REQUEST_OBJECT_ID_BITS) >> (64 - YAKSI_REQUEST_OBJECT_ID_BITS))
```
shintaro-iwasaki, line 67: Ditto.

src/util/yaksu_handle_pool.c
```diff
@@ -38,12 +38,18 @@ typedef struct handle {
...
     yaksu_handle_t next_handle; /* next brand-new handle (never allocated) */
     handle_s *free_handles;     /* list of handles that were allocated, but later freed */
     handle_s *used_handles;     /* hashmap of handles in use */
+
+    /* the first few handles are stored in a direct access cache, so
+     * we can get to these objects without having to grab a lock */
+    handle_s *handle_cache[HANDLE_CACHE_SIZE];
 } handle_pool_s;
 
 static pthread_mutex_t global_mutex = PTHREAD_MUTEX_INITIALIZER;
```
shintaro-iwasaki, line 55: It is not directly related to this PR, but do you need this global mutex?
In other words, if `global_mutex` is protecting global variables modified in macros, I should miss something important for reviewing this PR.

shintaro-iwasaki (commented):
The implementation looks correct in terms of inter-thread synchronization and correctness. A few minor comments.

src/frontend/include/yaksi.h
pavanbalaji: Fixed.

src/frontend/include/yaksi.h
pavanbalaji: Fixed.

src/util/yaksu_handle_pool.c
pavanbalaji: The global mutex is for pool creation.  Right now we only have two handle pools -- type handles and request handles.  So you are right that we don't need a global mutex.  But in the next PR we would have more pools (e.g., when we add multiple contexts).

pavanbalaji:
@shintaro-iwasaki  This is ready for your review again.

shintaro-iwasaki (approved):
Thank you! Looks good to me.

Issue #114, opened 2020-09-04
Lazy initialization of CUDA devices


shintaro-iwasaki:
At present, all CUDA devices including those not used by the main compute kernel are obtained and set up on initialization. On a machine that limits access of GPU resources to exclusive (i.e., only one process can access to GPU, the default setting of OLCF Summit), one process occupies all the CUDA devices, so creating multiple processes causes `all CUDA-capable devices are busy or unavailable` error in `yaksuri_cuda_init_hooks.c`.

There are several workarounds (e.g., setting `CUDA_VISIBLE_DEVICES`, changing the GPU resource access mode), but lazy initialization of GPU devices should be the most promising.

pavanbalaji:
I investigated this issue a bit.  I'm not sure if lazy initialization is the right solution here.  This error tends to happen when multiple processes are set to the same device.  Without `CUDA_VISIBLE_DEVICES` set, we have no way of avoiding it, even if we do lazy initialization because the access to a device can conflict even if such access is delayed.

The only practical way to use this would be with `CUDA_VISIBLE_DEVICES`, which is what the CUDA documentation seems to suggest for multi-process systems anyway.

PR #117, opened 2020-10-19
cuda: add device lazy initialization option

huanghua1994:
For systems without NVIDIA Multi-Process Service (MPS) support,
allows yaska to initialize and hook to the current CUDA device only,
and avoids unnecessary cudaSetDevice() from replacing CUDA
driver context. This lazy initialization is controlled by the environment
variable YAKSA_LAZY_INIT_DEVICE.

#114 

Signed-off-by: Huang, Hua <huangh223@gatech.edu>

## Author Checklist

### DIFF
--- a/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
+++ b/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
@@ -7,6 +7,7 @@
 #include "yaksuri_cudai.h"
 #include <assert.h>
 #include <string.h>
+#include <stdlib.h>
 #include <cuda.h>
 #include <cuda_runtime_api.h>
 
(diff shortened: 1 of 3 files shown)


raffenet:
I cleaned up the whitespace issue and rebased this on latest `main` branch. There was a power issue yesterday that keep CI jobs from running, but they should get kicked off now.

raffenet:
test:yaksa/warnings
test:yaksa/review

raffenet:
test:yaksa/review

pavanbalaji:
@huanghua1994  I'm not sure lazy initialization is needed.  I was told by the NVIDIA folks that creating the stream itself doesn't waste any resources, unless it is used.

huanghua1994:
@pavanbalaji Yes, creating the stream itself doesn't waste any resources unless it is used. The problem comes from the multi-process service (MPS). If multiple MPI processes are running on one node, all MPI processes will try to hook all GPUs on this node, which will crash the program when MPS is disabled. On the NERSC Cori GPU partition, we encountered this issue. I think lazy initialization allows the application to control the access of GPUs.

pavanbalaji:
> @pavanbalaji Yes, creating the stream itself doesn't waste any resources unless it is used. The problem comes from the multi-process service (MPS). If multiple MPI processes are running on one node, all MPI processes will try to hook all GPUs on this node, which will crash the program when MPS is disabled. On the NERSC Cori GPU partition, we encountered this issue. I think lazy initialization allows the application to control the access of GPUs.

This has already been fixed in Yaksa.  Lazy initialization was considered as a possible solution, but we decided that it doesn't quite fix this issue because even with lazy initialization two processes might end up triggering the GPUs at the same time.  So we went with an alternate solution in https://github.com/pmodels/yaksa/pull/123

huanghua1994:
Thanks for the clarification.

PR #120, opened 2020-10-23
Fix incorrect buffer checking

pavanbalaji:
## Pull Request Description

Fix incorrect buffer checking

## Author Checklist

### DIFF
--- a/src/backend/cuda/include/yaksuri_cudai.h
+++ b/src/backend/cuda/include/yaksuri_cudai.h
@@ -75,9 +75,9 @@ int yaksuri_cudai_md_alloc(yaksi_type_s * type);
 int yaksuri_cudai_populate_pupfns(yaksi_type_s * type);
 
 int yaksuri_cudai_ipack(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s * type,
-                        yaksi_info_s * info, yaksi_request_s * request, int target);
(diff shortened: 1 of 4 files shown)


yfguo (approved):

pavanbalaji:
Waiting for the CUDA tests on the JLSE machines to complete before merging this in.

zhenggb72:
Just looked at this PR, I see the yaksi_request_s argument to  yaksuri_*_ipack is removed. Can we undo this change?
It may still be useful to access the pointer type, and also we need a place to store immediate command queues (per yaksa_ipack call).

pavanbalaji:
> Just looked at this PR, I see the yaksi_request_s argument to yaksuri_*_ipack is removed. Can we undo this change? It may still be useful to access the pointer type, and also we need a place to store immediate command queues (per yaksa_ipack call). Actually this is not enough. I think we still need yaksuri_cudai_event_s to be passed to backend pack/unpack to store the command queues. We will need to free command queues inside yaksuri_cudai_event_destroy(void *event_)

@zhenggb72  Do you really need the event argument, or is the request structure what you are really looking for?  The request structure is generally meant to store such information.  Right now, we don't pass the request structure down to the driver layer because CUDA doesn't need it.  But, if ZE needs it, we can certainly pass it (and ignore it for CUDA).

PR #123, opened 2020-10-31
backend/cuda: disable CUDA in exclusive mode

pavanbalaji:
## Pull Request Description

If CUDA is setup in exclusive compute mode and the user did not filter the devices using CUDA_VISIBLE_DEVICES, disable CUDA support in Yaksa.

## Author Checklist

### DIFF
--- a/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
+++ b/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
@@ -113,6 +113,35 @@ int yaksuri_cuda_init_hook(yaksur_gpudriver_hooks_s ** hooks)
     cerr = cudaGetDeviceCount(&yaksuri_cudai_global.ndevices);
     YAKSURI_CUDAI_CUDA_ERR_CHKANDJUMP(cerr, rc, fn_fail);
 
+    if (getenv("CUDA_VISIBLE_DEVICES") == NULL) {
+        /* user did not do any filtering for us; if any of the devices
+         * is in exclusive mode, disable GPU support to avoid
+         * incorrect device sharing */
+        bool excl = false;
+        for (int i = 0; i < yaksuri_cudai_global.ndevices; i++) {
+            struct cudaDeviceProp prop;
+
+            cerr = cudaGetDeviceProperties(&prop, i);
+            YAKSURI_CUDAI_CUDA_ERR_CHKANDJUMP(cerr, rc, fn_fail);
+
+            if (prop.computeMode != cudaComputeModeDefault) {
+                excl = true;
+                break;
+            }
+        }
+
+        if (excl == true) {
+            fprintf(stderr, "[yaksa] ====> Disabling CUDA support <====\n");
+            fprintf(stderr,
+                    "[yaksa] CUDA is setup in exclusive compute mode, but CUDA_VISIBLE_DEVICES is not set\n");
+            fprintf(stderr,
+                    "[yaksa] You can silence this warning by setting CUDA_VISIBLE_DEVICES\n");
+            fflush(stderr);
+            *hooks = NULL;
+            goto fn_exit;
+        }
+    }
+
     yaksuri_cudai_global.stream = (cudaStream_t *)
         malloc(yaksuri_cudai_global.ndevices * sizeof(cudaStream_t));
 


pavanbalaji:
One concern with this PR is that, if CUDA is setup in exclusive mode, we would silently disable CUDA support inside Yaksa.  In this case, if the user passes a GPU buffer to Yaksa, it would assume that it's a CPU buffer and try to access it directly with load/stores and result in a segmentation fault.  Obviously, that's not ideal as it won't give a clear error message to the user.

What's the right solution here?  Should we throw an error message if CUDA is setup in exclusive mode and `CUDA_VISIBLE_DEVICES` is not set?  Such a warning could be incorrect in cases where the following three conditions are true: (1) CUDA is in exclusive mode, (2) the user did not set `CUDA_VISIBLE_DEVICES` and (2) there is only one process per node.

shintaro-iwasaki:
I have not tried, but the condition "there is only one process per node" might be too restrictive.  For example, If one node has multiple GPUs and the application partitions GPU resources per MPI process properly via a resource manager (e.g., on Summit), theoretically, it should work without `CUDA_VISIBLE_DEVICES`.

Regardless of whether Yaksa automatically disables CUDA or throws an error, it would be great if Yaksa/MPICH could tell users how to fix it (e.g., by partitioning GPU resources, setting `CUDA_VISIBLE_DEVICES`, or allowing some parameters to Yaksa to tell that the application uses only specific GPUs).

pavanbalaji:
> I have not tried, but the condition "there is only one process per node" might be too restrictive.

Of course.  We are not asking users to do that.  I'm simply stating that, if the user runs their application in that mode (e.g., MPI+OpenMP application), and Yaksa disables GPU support, then it might be counterintuitive.

> Regardless of whether Yaksa automatically disables CUDA or throws an error, it would be great if Yaksa/MPICH could tell users how to fix it (e.g., by partitioning GPU resources, setting `CUDA_VISIBLE_DEVICES`, or allowing some parameters to Yaksa to tell that the application uses only specific GPUs).

Yaksa looks for the `CUDA_VISIBLE_DEVICES` environment variable when in exclusive mode, even if the user can guarantee that there will not be any conflict (e.g., there is only one process on that node).

src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
```diff
@@ -113,6 +113,29 @@ int yaksuri_cuda_init_hook(yaksur_gpudriver_hooks_s ** hooks)
     cerr = cudaGetDeviceCount(&yaksuri_cudai_global.ndevices);
     YAKSURI_CUDAI_CUDA_ERR_CHKANDJUMP(cerr, rc, fn_fail);
 
+    if (getenv("CUDA_VISIBLE_DEVICES") == NULL) {
```
shintaro-iwasaki, line 116: The logic looks good to me to quickly fix the current runtime error. There seems no way to programmatically check the `CUDA_VISIBLE_DEVICES` property via CUDA API.
My concern is that this mechanism by default silently disables CUDA support. It would be user-friendly if it shows some hints (as warning if it is not an error) to enable it because I believe most users who use CUDA-enabled MPICH expect that this CUDA support is enabled.

pavanbalaji: Yeah, I was considering that, but it would throw a warning even in the correct case that I mentioned above.  I don't like incorrect warnings, but I guess I also don't like silently disabling something without telling the user.  Maybe I can throw a warning and ask the user to set `CUDA_VISIBLE_DEVICES` to disable the warning.  Let me add that fix.

pavanbalaji: @shintaro-iwasaki  Is this version better?

shintaro-iwasaki: Is setting CUDA_VISIBLE_DEVICES silence this warning? I believe users need to set CUDA_VISIBLE_DEVICES properly to enable CUDA support.

pavanbalaji: It disables this warning too.  For example, one can set `CUDA_VISIBLE_DEVICES=-1` to disable this warning, but still not get CUDA support.  So I think the description is correct.

yfguo (approved):

src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
shintaro-iwasaki: I got it. Thanks for clarification.

shintaro-iwasaki (approved):

PR #130, opened 2020-11-06
Allow negative extents in Yaksa

pavanbalaji:
## Pull Request Description

MPI allows negative extents, but did not allow it.  This PR adds that functionality to Yaksa.

## Author Checklist


yfguo (approved):
Look good to me!

test/dtpools/src/dtpools_attr.c
```diff
@@ -9,13 +9,13 @@
 #include <inttypes.h>
 
 #define VALUE_FITS_IN_INT(val) ((val) <= INT_MAX)
-#define VALUE_FITS_IN_UINTPTR_T(val) ((val) <= (((uint64_t) 1 << (sizeof(uintptr_t) * 8 - 1)) - 1))
+#define VALUE_FITS_IN_INTPTR_T(val) (((val) * 2) <= (((uint64_t) 1 << (sizeof(uintptr_t) * 8 - 1)) - 1))
```
hzhou, line 12: I don't this is correct. If `val` is a signed int, `val * 2` may overflow -- in fact, `val` itself may already overflow -- and become negative, which will return `true` here but apparently "does not fit", right?

pavanbalaji: `val` is `uint64_t`.  So, the assumption in the code is that `uint64_t` is large enough to accommodate it.  The problematic type would be `uintptr_t` which can be 32 bits.

test/dtpools/src/dtpools_attr.c
```diff
@@ -58,7 +58,7 @@ static int construct_contig(DTP_pool_s dtp, int attr_tree_depth, DTPI_Attr_s * a
 
         count /= attr->u.contig.blklen;
 
-        if (VALUE_FITS_IN_UINTPTR_T(extent * count))
+        if (VALUE_FITS_IN_INTPTR_T(extent * count))
```
hzhou, line 61: Following up on your reply (https://github.com/pmodels/yaksa/pull/130/commits/5c4c84e0ebdf7636cb26293dbd211307aa87a535#r519047604), both `extent` and `count` is signed integer now, right?

pavanbalaji: `extent` is `int64_t`.  This macro is checking if it fits in `intptr_t`.

hzhou: Does `INTPTR_T_MAX * 2` fit inside `intptr_t`? If `INTPTR_T_MAX * 2` overflow to negative, the macro will retrun `true`, right?

pavanbalaji: Oh, I don't know what that `* 2` is for.  It should simply compare `val`.

pavanbalaji: `If INTPTR_T_MAX * count overflow to negative`

Why is this happening?  It's `int64_t`, not `intptr_t`.

hzhou: What if `int64_t` and `intptr_t` is both `long int`?

pavanbalaji: Then this:

> val is int64_t. So, the assumption in the code is that int64_t is large enough to accommodate it. The problematic type would be intptr_t which can be 32 bits.

hzhou: I see.

PR #135, opened 2020-11-08
Generalize assignment code

pavanbalaji:
## Pull Request Description

Attempt to generalize the assignment code inside pack/unpack to allow for cases where the displacement of buffers passed by users does not directly align to the basic type size.

## Author Checklist

### DIFF
--- a/src/backend/cuda/genpup.py
+++ b/src/backend/cuda/genpup.py
@@ -35,14 +35,14 @@ def hvector(suffix, dtp, b, last):
     global s
     global idx
     global need_extent
-    yutils.display(OUTFILE, "intptr_t stride%d = %s->u.hvector.stride / sizeof(%s);\n" % (suffix, dtp, b))
+    yutils.display(OUTFILE, "intptr_t stride%d = %s->u.hvector.stride;\n" % (suffix, dtp))
     if (need_extent == True):
-        yutils.display(OUTFILE, "uintptr_t extent%d = %s->extent / sizeof(%s);\n" % (suffix, dtp, b))
+        yutils.display(OUTFILE, "uintptr_t extent%d = %s->extent;\n" % (suffix, dtp))
     if (last != 1):
         s += " + x%d * stride%d + x%d * extent%d" % (idx, suffix, idx + 1, suffix + 1)
         need_extent = True
     else:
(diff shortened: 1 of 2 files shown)


pavanbalaji:
This PR is not fully ready yet.  The code is written in a way to try to minimize performance impact, but the actual impact is not fully analyzed yet.

pavanbalaji:
There are three types of packing kernels that need to be considered:

1. Array-style assignments.  This is the original version of the code in yaksa.  It essentially looks like a multiple level for loop followed by an array assignment from the source buffer to the destination buffer.  This format is the easiest for the compiler to optimize because it implicitly knows that the data is correctly aligned to the correct datatype.  Also, arrays are easier for the compiler to vectorize.  This version assumes that (1) all buffers are correctly aligned and that (2) the buffers are essentially arrays of the basic type -- so the displacements have to be multiple of the type size.

2. Pointer-offset style assignment.  This is the version of the code created by this PR currently.  The generated assembly code right now seems to be identical to the array-format assignment, so we might not lose performance.  But it'll likely conflict with @shintaro-iwasaki's work on vectorization and would likely hurt performance when vectorization is enabled, at least if the compiler cannot analyze the code to convert it to an array format.  This version retains the first assumption above, but drops the second assumption.

3. Fully-generic types.  This version assumes no alignment.  So we cannot perform a direct assignment of types.  We'd need to convert the code to either use memcpy or a loop of assignments for byte-sized elements.  This version would likely lose a lot more performance.

pavanbalaji:
Based on initial evaluation, the loss of vectorization is unclear with this PR.  We obviously need to investigate it further, but for now, this PR should be OK.

This still does not support completely arbitrary (no alignment) datatypes.  But it is hard to see whether that is needed.

shintaro-iwasaki (approved):
> Based on initial evaluation, the loss of vectorization is unclear with this PR. We obviously need to investigate it further, but for now, this PR should be OK.

Yes I agree. I approved this PR. If it passes all the tests, it should be good to merge.

pavanbalaji:
Checking on CUDA systems too, to confirm before merging.

pavanbalaji:
Unclear why the configs test failed.  Retrying.

test:yaksa/review/configs

pavanbalaji:
test:yaksa/review

PR #140, opened 2020-11-11
ZE backend for Intel GPUs

zhenggb72:
## Author Checklist


Makefile.am
```diff
@@ -34,6 +34,8 @@ else
 libyaksa_la_LDFLAGS = -version-info @libyaksa_so_version@
 endif !EMBEDDED_BUILD
 
+ze_native_TARGET = @enable_ze_native@
```
pavanbalaji, line 37: Why is this sitting in the top-level Makefile?

src/backend/gencomm.py
```diff
@@ -112,8 +112,13 @@ def populate_pupfns(pup_max_nesting, backend, blklens, builtin_types, builtin_ma
     yutils.display(OUTFILE, "yaksuri_%si_type_s *%s = (yaksuri_%si_type_s *) type->backend.%s.priv;\n" \
                    % (backend, backend, backend, backend))
     yutils.display(OUTFILE, "\n")
+    yutils.display(OUTFILE, "#ifdef KERNEL_NULL\n")
+    yutils.display(OUTFILE, "%s->pack = KERNEL_NULL;\n" % backend)
+    yutils.display(OUTFILE, "%s->unpack = KERNEL_NULL;\n" % backend)
+    yutils.display(OUTFILE, "#else\n")
```
pavanbalaji, line 118: This seems messy.  `KERNEL_NULL` is a ze-specific macro, which is being checked for in `gencomm`, which is not ze-specific.  It might be better to replace this with one extra level of indirection.  For example, each driver can maintain an array of kernels (or pointers to kernels) and the pack/unpack variables can simply maintain indices within that array.  That way, we can use index `-1` to mean that this type does not have a pack/unpack kernel.

test/pack/pack.c
```diff
@@ -16,6 +16,11 @@
 #include <cuda_runtime_api.h>
 #endif /* HAVE_CUDA */
 
+#ifdef HAVE_ZE
+#include "level_zero/ze_api.h"
+
```
pavanbalaji, line 21: Extra empty line.

test/pack/pack.c
```diff
@@ -60,18 +65,71 @@ typedef enum {
...
 static int ndevices = -1;
 #endif
 static int device_id = 0;
 static int device_stride = 0;
 
+#ifdef HAVE_ZE
+static ze_driver_handle_t global_ze_driver_handle;
+static ze_device_handle_t *global_devices = NULL;
+static ze_context_handle_t ze_context;
+#endif
```
pavanbalaji, line 78: The code is getting very messy.  It might be better to move the CUDA and ZE pieces into separate files, and have `pack.c` simply access them through some interface.

test/pack/pack.c
```diff
@@ -235,12 +417,10 @@ void *runtest(void *arg)
         rc = yaksa_type_get_size(dobj.DTP_datatype, &dsize);
         assert(rc == YAKSA_SUCCESS);
 
-
```
pavanbalaji, line 238: Unnecessary change.

test/pack/pack.c
```diff
@@ -235,12 +417,10 @@ void *runtest(void *arg)
...
         /* the source and destination objects should have the same
          * signature */
         assert(ssize * sobj.DTP_type_count == dsize * dobj.DTP_type_count);
 
-
```
pavanbalaji, line 243: Unnecessary change.

pavanbalaji (changes requested):

Makefile.am
zhenggb72: This seems to be the only am file there, variable ze_natibe_Target is an am variable. Adding to the ze/Makefile.mk did not work. But I may miss something. let me know if you have better solution.

src/backend/gencomm.py
zhenggb72: I don't see why it is ze-specific macro. Any backend can define KERNEL_NULL. To clean it up, I can use KERNEL_NULL in this script for all backends. And define KERNEL_NULL to be "NULL" for cuda,   and "-1" for ze.

pavanbalaji: It's defined in `yaksuri_zei.h`.  Ergo, it's ze-specific.  Yes, you can fix it in multiple ways.  But the current code is not one of those.  It needs to be changed.  Also, if you want to make it generic, we'll need a namespace to it: `YAKSURI

Makefile.am
pavanbalaji: I'll need to experiment with it a bit to figure it out.  But this does not seem like the right solution.

src/backend/gencomm.py
zhenggb72: Ok, I will fix it by naming it YAKSRI_KERNEL_NULL, and define it in each backend.

Makefile.am
pavanbalaji: I looked at this a bit, and I'm not really sure what you are trying to do here.  Why are you defining `ze_native_TARGET` instead of directly using `@enable_ze_native@` inside `src/backend/ze/pup/Makefile.mk`?  The name `@enable_ze_native@` is a bit weird and probably should be changed, but functionality wise, this seems like a very round-about approach, unless I'm missing something here.

test/pack/pack.c
zhenggb72: fixed

Makefile.am
zhenggb72: fixed

test/pack/pack.c
zhenggb72: done

zhenggb72:
I will look into failures on cuda

test/pack/pack-common.c
```diff
@@ -0,0 +1,94 @@
...
+#elif defined(HAVE_ZE)
+    } else {
+        pack_ze_alloc_mem(size, type, hostbuf, devicebuf);
+    }
+#else
+    } else {
+        fprintf(stderr, "ERROR: no GPU device is supported\n");
+        exit(1);
+    }
+#endif
```
pavanbalaji, line 53: This will not compile if yaksa finds both CUDA and ZE libraries on the system.  How about the following style instead:

```
if (...) {
   /* some stuff */
#Ifdef defined(HAVE_CUDA)
} else {
   /* some stuff */
#if defined(HAVE_ZE)
} else {
   /* some stuff */
#endif
} else {
   /* some stuff */
}
```

zhenggb72: This does not work if either HAVE_CUDA and HAVE_ZE is defined, because it will have two "else" branches.

pavanbalaji: Right.  Something like this might work.  If not, some experimentation can help you figure out what will work:

```
if (...) {
   /* some stuff */
} else {
#ifdef defined(HAVE_CUDA)
   /* some stuff */
#elif defined(HAVE_ZE)
   /* some stuff */
#else
   /* some stuff */
#endif
}
```

zhenggb72: changed

pavanbalaji (approved):

pavanbalaji:
test:yaksa/whitespace

pavanbalaji:
test:yaksa/whitespace

pavanbalaji:
I've merged this PR.  Thank you for putting it together, @zhenggb72

PR #144, opened 2020-11-15
Fixes to the pack test

pavanbalaji:
## Pull Request Description

This PR includes several fixes including corrections to the multi-threaded code as well as reduction to the number of tests.

## Author Checklist


test/pack/pack-ze.c
```diff
@@ -17,11 +17,27 @@
...
+    assert(ret == ZE_RESULT_SUCCESS);
+    assert(driver_count > 0);
+
+    all_drivers = malloc(driver_count * sizeof(ze_driver_handle_t));
+    assert(all_drivers != NULL);
+    ret = zeDriverGet(&driver_count, all_drivers);
+    assert(ret == ZE_RESULT_SUCCESS);
+
+    uint32_t device_count = 0;
+    ret = zeDeviceGet(all_drivers[0], &device_count, NULL);
```
zhenggb72, line 37: code is correct, but why not store the device_count when calling pack_ze_init_devices?
or change pack_ze_init_device to return the count (so one less function to maintain).

pavanbalaji: It seems weird to call `init_devices` when all we need is the number of devices.  I prefer simpler functions that do one thing, rather than large functions that do many things.

zhenggb72:
Does this mean now pack test is always a multi-device test when possible?

pavanbalaji:
> Does this mean now pack test is always a multi-device test when possible?

Correct.

zhenggb72:
That's ok, although it is less flexible than before where one can narrow the test cases.

pavanbalaji:
> That's ok, although it is less flexible than before where one can narrow the test cases.

That's correct.  But it's a compromise in order to reduce the number of tests.

test/pack/pack-ze.c
zhenggb72: I meant we call init_device function which returns the device count, so we don't need a separate function to get the count.

pavanbalaji: Then, the function should not do anything more than getting the number of devices.  For example, it cannot create any contexts like the current code does.

Btw, I just realized that the ZE version of the function is completely wrong.  :-).  It's not returning anything.  :-). I'll fix it in a bit, but I might need you to test it out.

pavanbalaji: Pushed my changes.  FYI, I don't want a function that does "init_devices_and_get_ndevices".  I prefer having two simpler functions instead: "init_devices" and "get_ndevices".  This allows the caller to call get_ndevices any number of times without having to worry about it.  On the other hand, "init_devices" clearly goes together with "finalize_devices" to create/free resources.

zhenggb72 (approved):

zhenggb72 (approved):

pavanbalaji:
test:yaksa/configs

pavanbalaji:
test:yaksa/review/configs

yfguo (approved):

PR #147, opened 2020-11-16
m4: Add -Wpedantic to strict flags

raffenet:
## Pull Request Description

Add -Wpedantic to catch violations of ISO C, such as returning a value
from a void function that have slipped into the code in the past.

## Author Checklist

### DIFF
--- a/m4/aclocal_cc.m4
+++ b/m4/aclocal_cc.m4
@@ -548,6 +548,7 @@ if test "$enable_strict_done" != "yes" ; then
         -Wpacked
         -Wshorten-64-to-32
         -diag-disable=all
+        -Wpedantic
     "
 
     enable_c89=no


raffenet:
Reverted #146 to confirm the new strict warnings check will catch the problematic code.

raffenet:
Void return warning was correctly showed with this PR. https://jenkins-pmrs.cels.anl.gov/job/yaksa-warnings/181/

Problem: the addition of this flag causes a number of warnings because of the inclusion of `cuda.h` (anonymous struct/union, invalid enumerator value).

raffenet:
> Problem: the addition of this flag causes a number of warnings because of the inclusion of `cuda.h` (anonymous struct/union, invalid enumerator value).

We can raise the C standard level to `c11` for CUDA builds to silence the anonymous struct/union warnings. The enum value may need to be fixed in `cuda.h`, though.

pavanbalaji:
> > Problem: the addition of this flag causes a number of warnings because of the inclusion of `cuda.h` (anonymous struct/union, invalid enumerator value).
> 
> We can raise the C standard level to `c11` for CUDA builds to silence the anonymous struct/union warnings. The enum value may need to be fixed in `cuda.h`, though.

I think adding `-Wpedantic` is more problematic than helpful.  I'd recommend simply dropping it.

pavanbalaji:
@raffenet   What's the plan for this?  Should this be closed?

raffenet:
I agree we should just drop it.

PR #150, opened 2020-11-17
backend: support managed buffer that is associated to a single device

zhenggb72:
Signed-off-by: Gengbin Zheng <gengbin.zheng@intel.com>

## Author Checklist

### DIFF
--- a/src/backend/src/yaksuri_progress.c
+++ b/src/backend/src/yaksuri_progress.c
@@ -896,7 +896,9 @@ int yaksuri_progress_enqueue(const void *inbuf, void *outbuf, uintptr_t count, y
             goto enqueue_subreq;
         }
 
-        if (request->backend.outattr.type == YAKSUR_PTR_TYPE__MANAGED) {
+        if (request->backend.outattr.type == YAKSUR_PTR_TYPE__MANAGED &&
+            (request->backend.outattr.device == -1 ||
+             request->backend.inattr.device == request->backend.outattr.device)) {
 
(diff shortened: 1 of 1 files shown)


src/backend/src/yaksur_pup.c
```diff
@@ -46,6 +46,13 @@ static int ipup(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s *
                 request->backend.outattr.type == YAKSUR_PTR_TYPE__GPU) {
                 reqpriv->gpudriver_id = id;
                 break;
+            } else if (request->backend.inattr.type == YAKSUR_PTR_TYPE__MANAGED &&
+                       request->backend.outattr.type == YAKSUR_PTR_TYPE__MANAGED &&
+                       request->backend.inattr.device != -1 &&
+                       request->backend.outattr.device != -1 &&
+                       request->backend.inattr.device != request->backend.outattr.device) {
+                reqpriv->gpudriver_id = id;
+                break;
```
pavanbalaji, line 55: I thought the intent of this PR was to handle cases where one buffer was on the GPU and the other was managed but could not migrate to the other GPU.  In such cases, since at least one of the buffers is on the GPU, the first "if" already handles this.

zhenggb72: ok, you are right. This case is not needed.

pavanbalaji: There is a bunch of code in this PR that is following the same incorrect logic.  This PR needs several fixes.

pavanbalaji (changes requested):

pavanbalaji (approved):

PR #169, opened 2021-01-14
infohint: add nogpu as a gpudriver_id hint

### LINKED ISSUE #168: Add an info hint key to disable gpu buffer query
When an application is using a CPU buffer, it is desirable to hint yaksa to avoid unnecessary GPU attribute queries. Currently, we have to supply per driver attribute which is clumsy when user simply wants to disable the GPU check.

hzhou:
## Pull Request Description

The purpose of this PR is add a way to tell yaksa the both `in_buf` and `out_buf` is cpu host memory. This is useful when yaksa is configured with GPU support but the application doesn't use gpu memory or the application knows when their buffer is entirely from CPU and thus doesn't need to pay the pointer attribute query overhead.

## Expected Impact

Fixes #168 

## Author Checklist

### DIFF
--- a/test/pack/pack.c
+++ b/test/pack/pack.c
@@ -49,6 +49,22 @@ static void swap_segments(uintptr_t * starts, uintptr_t * lengths, int x, int y)
     lengths[y] = tmp;
 }
 
+static void host_only_get_ptr_attr(yaksa_info_t * info)
+{
+    static int count = 0;
+    if ((++count) % 2 == 0) {
+        int rc;
+
+        rc = yaksa_info_create(info);
+        assert(rc == YAKSA_SUCCESS);
+
+        rc = yaksa_info_keyval_append(*info, "yaksa_gpu_driver", "nogpu", strlen("nogpu"));
+        assert(rc == YAKSA_SUCCESS);
+    } else {
+        *info = NULL;
+    }
+}
+
 char typestr[MAX_DTP_BASESTRLEN + 1] = { 0 };
 
 int seed = -1;
@@ -228,8 +244,16 @@ void *runtest(void *arg)
         pack_alloc_mem(tbuf_devid, tbufsize, tbuf_memtype, &tbuf_h, &tbuf_d);
 
         yaksa_info_t pack_info, unpack_info;
-        pack_get_ptr_attr(sbuf_d + sobj.DTP_buf_offset, tbuf_d, &pack_info);
-        pack_get_ptr_attr(tbuf_d, dbuf_d + dobj.DTP_buf_offset, &unpack_info);
+        if (sbuf_memtype != MEM_TYPE__DEVICE && tbuf_memtype != MEM_TYPE__DEVICE) {
+            host_only_get_ptr_attr(&pack_info);
+        } else {
+            pack_get_ptr_attr(sbuf_d + sobj.DTP_buf_offset, tbuf_d, &pack_info);
+        }
+        if (tbuf_memtype != MEM_TYPE__DEVICE && dbuf_memtype != MEM_TYPE__DEVICE) {
+            host_only_get_ptr_attr(&unpack_info);
+        } else {
+            pack_get_ptr_attr(tbuf_d, dbuf_d + dobj.DTP_buf_offset, &unpack_info);
+        }
 
         for (int j = 0; j < segments; j++) {
             uintptr_t actual_pack_bytes;
--- a/test/pack/pack.c
+++ b/test/pack/pack.c
@@ -49,10 +49,9 @@ static void swap_segments(uintptr_t * starts, uintptr_t * lengths, int x, int y)
     lengths[y] = tmp;
 }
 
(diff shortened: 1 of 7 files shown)


hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
@pavanbalaji The PR is ready for your review. There is 1 failure in the newly created gpu testing, see https://github.com/pmodels/yaksa/pull/169#issuecomment-760615322, I am not sure it is related to this PR. Maybe you can help to decide. Meanwhile, I'll re-run the test to confirm.

hzhou:
tets:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
@pavanbalaji Tests are clean. Ready for your review.

test/pack/pack.c
```diff
@@ -228,8 +244,16 @@ void *runtest(void *arg)
...
+        if (sbuf_memtype != MEM_TYPE__DEVICE && tbuf_memtype != MEM_TYPE__DEVICE) {
+            host_only_get_ptr_attr(&pack_info);
+        } else {
+            pack_get_ptr_attr(sbuf_d + sobj.DTP_buf_offset, tbuf_d, &pack_info);
+        }
+        if (tbuf_memtype != MEM_TYPE__DEVICE && dbuf_memtype != MEM_TYPE__DEVICE) {
+            host_only_get_ptr_attr(&unpack_info);
+        } else {
+            pack_get_ptr_attr(tbuf_d, dbuf_d + dobj.DTP_buf_offset, &unpack_info);
+        }
```
pavanbalaji, line 256: This patch is forcing all host-only testing to use this new info hint.  This will cause us to lose all testing for the cases where this info hint is not provided.

pavanbalaji (changes requested):

test/pack/pack.c
hzhou: It only adds the info hint to half of the cases, similar to the gpu tests. Refer to line 55 above.

test/pack/pack.c
```diff
@@ -49,6 +49,22 @@ static void swap_segments(uintptr_t * starts, uintptr_t * lengths, int x, int y)
     lengths[y] = tmp;
 }
 
+static void host_only_get_ptr_attr(yaksa_info_t * info)
+{
+    static int count = 0;
+    if ((++count) % 2 == 0) {
```
pavanbalaji, line 55: Is this thread-safe?

hzhou: Not really. But the consequence is just less than perfect half by half testing.  The same  mechanism is used in `test/pack/pack-cuda.c`. We can make it atomic (do we have atomic library for the testing?) but maybe it isn't worth the trouble.

pavanbalaji: No, randomness is not desirable in this case.  It's better to make both cases deterministic.  If we don't have an atomic library, we could use a lock or decide on the info hint based on the iteration number.

hzhou: Okay

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
@pavanbalaji ready for your review again

pavanbalaji:
How did the failures disappear?  Did you simply rerun the tests till they disappeared?  That's not a good model.  It only means that the errors are intermittent and are being ignored.

hzhou:
> 
> 
> How did the failures disappear? Did you simply rerun the tests till they disappeared? That's not a good model. It only means that the errors are intermittent and are being ignored.

I did record them down. They seem to be intermittent gpu errors. I won't have time to debug them. I'll have to check them later.

pavanbalaji:
If they are happening with `main` too, then we should at least have an issue created for it.  If it's not happening with `main`, then it's a problem with this PR and it cannot be merged till that's fixed.

hzhou:
> 
> 
> If they are happening with `main` too, then we should at least have an issue created for it. If it's not happening with `main`, then it's a problem with this PR and it cannot be merged till that's fixed.

Let me open a dummy PR to check that

pavanbalaji:
I meant in the nightly tests.  You don't need a dummy PR.

hzhou:
> 
> 
> I meant in the nightly tests. You don't need a dummy PR.


> 
> 
> I meant in the nightly tests. You don't need a dummy PR.

Sure. Let me create one.

hzhou:
@pavanbalaji The cuda memory issue was reproduced in nightly test (https://jenkins-pmrs.cels.anl.gov/view/yaksa/job/yaksa-nightly-gpu/lastCompletedBuild/testReport/) and tracked in issue https://github.com/pmodels/yaksa/issues/170

hzhou:
@pavanbalaji Is there any remaining issues with this PR?

pavanbalaji (approved):

PR #171, opened 2021-01-19
type: change the type creation API to use intptr_t

hzhou:
## Pull Request Description

Modify all yaksa type creation function, e.g.  `yaksa_type_create_contig_x`, `yaksa_type_create_vector_x`, etc. to use `intptr_t` typed arguments in stead of `int`. This really only affects:
* `yaksa_type_create_indexed_block`
* `yaksa_type_create_indexed`
* `yaksa_type_create_hindexed`
* `yaksa_type_create_struct`
* `yaksa_type_create_subarray`
, because they now use `intptr_t` arrays rather than `int` array. For all scalar parameters, it should be automatically converted to the larger integer type gracefully by the compiler.

## Author Checklist

### DIFF
--- a/examples/fuf.c
+++ b/examples/fuf.c
@@ -17,7 +17,7 @@ int main()
     int pack_buf[SIZE];
     int unpack_buf[SIZE];
     yaksa_type_t indexed_block;
-    int array_of_displacements[ROWS] = {
+    intptr_t array_of_displacements[ROWS] = {
         4, 12, 20, 28,
         32, 40, 48, 56
     };
--- a/examples/hindexed.c
+++ b/examples/hindexed.c
@@ -14,7 +14,7 @@ int main()
     int pack_buf[SIZE];
     int unpack_buf[SIZE];
     yaksa_type_t hindexed;
-    int array_of_blocklengths[ROWS - 1] = {
+    intptr_t array_of_blocklengths[ROWS - 1] = {
         1, 2, 2, 4, 4, 4, 4
     };
     intptr_t array_of_displacements[ROWS - 1] = {
--- a/examples/indexed.c
+++ b/examples/indexed.c
@@ -14,10 +14,10 @@ int main()
     int pack_buf[SIZE];
     int unpack_buf[SIZE];
     yaksa_type_t indexed;
-    int array_of_blocklengths[ROWS - 1] = {
+    intptr_t array_of_blocklengths[ROWS - 1] = {
         1, 2, 2, 4, 4, 4, 4
     };
-    int array_of_displacements[ROWS - 1] = {
+    intptr_t array_of_displacements[ROWS - 1] = {
         9,
         18, 26,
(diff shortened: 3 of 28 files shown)


hzhou:
@pavanbalaji This PR is ready for your review. Currently, I simply added `_x` versions type creation routines. If you prefer, I can easily add another commit rename and remove the extra interface.

pavanbalaji:
I think it's better to just replace the old functions in this case.

pavanbalaji:
Also, why `intptr_t` instead of `uintptr_t`?  Count and blocklength cannot be negative.

hzhou:
> 
> 
> Also, why `intptr_t` instead of `uintptr_t`? Count and blocklength cannot be negative.

Mostly compatibility with `MPI_Aint` -- the less sign/unsign coercion, the less bugs.

pavanbalaji:
> > Also, why `intptr_t` instead of `uintptr_t`? Count and blocklength cannot be negative.
> 
> Mostly compatibility with `MPI_Aint` -- the less sign/unsign coercion, the less bugs.

We are breaking MPI compatibility anyway by moving it to `intptr_t` instead of `int`.  So if we are making that change, it's better to pick the right type.

hzhou:
> We are breaking MPI compatibility anyway by moving it to `intptr_t` instead of `int`. So if we are making that change, it's better to pick the right type.

Well, I always believe that signed int is the right type, where the only con being losing the extremely large upper half range. The pro being safety in subtraction and comparisons (which are subtractions). Converting `unsigned` to `signed` is nearly always safe. The other way is very dangerous. As a library sitting on the acceptance side, I think we should always use signed integers (unless it is not an arithmetic number).

Also here having signed for displacement and unsigned for lengths and having them involved in arithmetics (which forces the conversion back and forth) seem to me a nightmare scenario).

pavanbalaji:
`uintptr_t` is more precise based on the semantics of those variables.  If you are particular about error cases, you can add additional error checking in the code.

hzhou:
> 
> 
> `uintptr_t` is more precise based on the semantics of those variables. If you are particular about error cases, you can add additional error checking in the code.

For example, a user calls your library:
```
int a = somevalue_that_become_negative;
your_library_func((uintptr_t) a);
```

The example shows that declaring the type as unsigned has no effect preventing user passing your negative value. If the library accepts signed integer, the library can tell that user passed in bad value and do error return. With unsigned type, there is no way to check, and it may well end up exhaust the system memory.

I don't think there is any benefit of using unsigned type here -- nothing precise about it.

hzhou:
The tests hang. Need fixing

hzhou:
Updated dtpools to use `intptr_t`. but tests still hangs. I might have to debug ...

hzhou:
test:yaksa/gpu

raffenet (approved):

hzhou:
test:yaksa/review/platforms

PR #172, opened 2021-01-24
init: skip gpu initialization with infohint

hzhou:
## Pull Request Description

Yaksa will query gpu devices during init. This query can have huge
latency especially when multiple processes are trying to query the device at
the same time. With CUDA, cudaGetDeviceCount and checking peer access
are the most suceptible operations.

This commit allows "nogpu" infohint to skip the gpu driver setup.

Skipping gpu driver setup leaves yaksuri_global.gpudriver[id].hooks
NULL, thus will skip driver finalize. If somehow there is access to gpu
driver during runtime, I guess it will simply segfault (due to NULL
hooks).

TODO: a better solution may be to allow lazy gpu initialization. This
can be helpful especially when not all process on the same node need
access GPU.

## Author Checklist

### DIFF
--- a/src/backend/src/yaksur_hooks.c
+++ b/src/backend/src/yaksur_hooks.c
@@ -65,13 +65,21 @@ static void free_fn(void *buf, void *state)
     assert(0);
 }
 
-int yaksur_init_hook(void)
+int yaksur_init_hook(yaksi_info_s * info)
 {
     int rc = YAKSA_SUCCESS;
     yaksuri_gpudriver_id_e id;
 
     rc = yaksuri_seq_init_hook();
     YAKSU_ERR_CHECK(rc, fn_fail);
+
+    if (info) {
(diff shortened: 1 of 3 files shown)


hzhou:
@pavanbalaji Would you be interested in reviewing this PR?

hzhou:
It is important to be able to skip yaksa gpu init. In MPICH's test suite there is a `darray_pack` test that runs with 72 process. Yaksa will try to `cudaStreamCreateWithFlags` for every device *every process*, and will hit following error:
```
not ok  - ./datatype/darray_pack 72
  ...
## Test output (expected 'No Errors'):
## CUDA Error (yaksuri_cuda_init_hook:src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c,162): out of memory
...
```

zhenggb72 (approved):
changes look good to me

yfguo (approved):
LGTM

PR #179, opened 2021-04-11
backend/cuda: refactoring complex pack/unpack protocol switch in progress enqueue and minor bug fix

minsii:
## Pull Request Description
The current pack/unpack protocol switch code in `yaksuri_progress_enqueue` is complex and not easy to read. This patch refactors the long if-else based switch logic. The basic approach is to separate each situation (e.g., d2d_pack, d2rh_pack...) into subroutines so that later developers can easily find the place to optimize.

## Author Checklist

### DIFF
--- a/src/backend/src/yaksuri_progress.c
+++ b/src/backend/src/yaksuri_progress.c
@@ -1065,6 +1065,42 @@ static int set_multichunk_subreq(const void *inbuf, void *outbuf, uintptr_t coun
     return rc;
 }
 
+/* Fast-path pack subroutine */
+static int singlechunk_pack(yaksuri_gpudriver_id_e id, int device, const void *inbuf, void *outbuf,
+                            uintptr_t count, yaksi_type_s * type, yaksi_info_s * info,
+                            yaksa_op_t op, yaksuri_subreq_s * subreq)
+{
+    int rc = YAKSA_SUCCESS;
+    subreq->kind = YAKSURI_SUBREQ_KIND__SINGLE_CHUNK;
+    rc = ipack(id, inbuf, outbuf, count, type, info, op, device);
+    YAKSU_ERR_CHECK(rc, fn_fail);
+
+    rc = event_record(id, device, &subreq->u.single.event);
+
+  fn_exit:
+    return rc;
+  fn_fail:
+    goto fn_exit;
+}
+
+/* Fast-path unpack subroutine */
+static int singlechunk_unpack(yaksuri_gpudriver_id_e id, int device, const void *inbuf,
+                              void *outbuf, uintptr_t count, yaksi_type_s * type,
+                              yaksi_info_s * info, yaksa_op_t op, yaksuri_subreq_s * subreq)
+{
+    int rc = YAKSA_SUCCESS;
+    subreq->kind = YAKSURI_SUBREQ_KIND__SINGLE_CHUNK;
+    rc = iunpack(id, inbuf, outbuf, count, type, info, op, device);
+    YAKSU_ERR_CHECK(rc, fn_fail);
+
+    rc = event_record(id, device, &subreq->u.single.event);
+
+  fn_exit:
+    return rc;
+  fn_fail:
+    goto fn_exit;
+}
+
 /* Subroutines to setup subreq for pack with different buffer types */
 static int set_subreq_pack_from_device(const void *inbuf, void *outbuf, uintptr_t count,
                                        yaksi_type_s * type, yaksi_info_s * info, yaksa_op_t op,
@@ -1098,11 +1134,8 @@ static int set_subreq_pack_from_device(const void *inbuf, void *outbuf, uintptr_
                 YAKSU_ERR_CHECK(rc, fn_fail);
             }
         } else if (request->backend.inattr.device == request->backend.outattr.device) {
-            subreq->kind = YAKSURI_SUBREQ_KIND__SINGLE_CHUNK;
-            rc = ipack(id, inbuf, outbuf, count, type, info, op, request->backend.inattr.device);
-            YAKSU_ERR_CHECK(rc, fn_fail);
-
-            rc = event_record(id, request->backend.inattr.device, &subreq->u.single.event);
+            rc = singlechunk_pack(id, request->backend.inattr.device, inbuf, outbuf, count,
+                                  type, info, op, subreq);
             YAKSU_ERR_CHECK(rc, fn_fail);
         } else {
             bool is_enabled;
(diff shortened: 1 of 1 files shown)


minsii:
test:yaksa/gpu

minsii:
@hzhou would you be able to review this PR? It is mostly refactoring for better readability. I am doing more fixes/optimizations  based on the refactored version which I will post as separate PRs.

minsii:
Just realized that the commit messages should be changed from `backend/cuda` to `backend`. No code change. Do we need test `ze`?

hzhou:
> 
> 
> Just realized that the commit messages should be changed from `backend/cuda` to `backend`. No code change. Do we need test `ze`?

Ideally, yes. Unfortunately, that is difficult and adds a significant burden, so unless there are explicit changes to ZE-speicific code, I think we can relax the testing requirement.

src/backend/src/yaksuri_progress.c
```diff
@@ -1065,333 +1065,439 @@ static int set_multichunk_subreq(const void *inbuf, void *outbuf, uintptr_t coun
     return rc;
 }
 
-int yaksuri_progress_enqueue(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s * type,
-                             yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request)
+/* Subroutines to setup subreq for pack with different buffer types */
+static int set_pack_d2_subreq(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s * type,
+                              yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request,
+                              yaksuri_request_s * reqpriv, uintptr_t threshold,
+                              yaksuri_subreq_s * subreq)
```
hzhou: `d` means "device", right? What does `2` refer to?

src/backend/src/yaksuri_progress.c
```diff
@@ -1065,6 +1065,42 @@ static int set_multichunk_subreq(const void *inbuf, void *outbuf, uintptr_t coun
     return rc;
 }
 
+/* Fast-path pack subroutine */
+static int singlechunk_pack(yaksuri_gpudriver_id_e id, int device, const void *inbuf, void *outbuf,
+                            uintptr_t count, yaksi_type_s * type, yaksi_info_s * info,
+                            yaksa_op_t op, yaksuri_subreq_s * subreq)
```
hzhou: The name need to reflect this is for same-device transfer (is it?).

hzhou (commented):
Generally looks good. Just a few minor comments

src/backend/src/yaksuri_progress.c
```diff
@@ -17,12 +17,12 @@
 static yaksuri_request_s *pending_reqs = NULL;
 static pthread_mutex_t progress_mutex = PTHREAD_MUTEX_INITIALIZER;
 
-static bool buf_is_aligned(const void *buf, yaksi_type_s * type)
+static inline bool buf_is_aligned(const void *buf, yaksi_type_s * type)
```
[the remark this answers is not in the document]
minsii: I will check the optimized build to make sure that these functions are inlined.

src/backend/src/yaksuri_progress.c
minsii: Yes, `d` means `device`. `2` refers to `to`. It tries to follow the naming of existing functions such as `pack_d2d_nop2p_acquire`

src/backend/src/yaksuri_progress.c
minsii: Not really. The current code only uses it for same-device fast path. But we will also use it to optimize the IPC and P2P-enabled cases in the upcoming optimization PR.

src/backend/src/yaksuri_progress.c
hzhou: I see. How about `set_subreq_pack_from_device`?

src/backend/src/yaksuri_progress.c
hzhou: okay

src/backend/src/yaksuri_progress.c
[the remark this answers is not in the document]
minsii: Confirmed that without `inline` the compiler can good job to inline those subroutines (tried gcc v7.5.0, assuming clang and icc can do better). Will remove the inline commit in the next push.

src/backend/src/yaksuri_progress.c
minsii: The proposed format may make some subroutine names very long, e.g.,`set_subreq_pack_from_registered_host`, `set_subreq_pack_from_device_to_device`, `set_subreq_pack_from_device_to_registered_host`. I think it is fine to use these abbreviation consistently in this file.

If you are OK with the abbreviation, I am not sure which is better between `set_pack_d2_subreq ` and `set_subreq_pack_d2`. What do you think?

src/backend/src/yaksuri_progress.c
minsii: An update for the naming of `singlechunk_pack`: after second read on this code, this routine is used only for `YAKSA_OP__REPLACE` in the current code. It is still unclear to me whether the same fast path can be used for other ops (e.g., MAX, MIN). But I guess it is OK to keep the name for now. Once we make optimization for non-REPLACE ops, we can adjust at that time.

minsii:
This push deletes the inline commit

src/backend/src/yaksuri_progress.c
hzhou: We are more used to `d2d`, `d2h`, etc. although a short comment above the definition to expand the abbreviation can be helpful.

`_d2` is just too cryptic, mostly because we don't use such often. When used, it often means different from the meaning here. One of the heuristics (from "Clean code") is the name needs to be pronounceable and can be put into a sentence. `d2` doesn't make sense in a sentence.

minsii: Agreed with your comment. But just the suggested name seems too long and a bit inconsistent with the other subroutines which use `d2d, d2h`. I guess the confusing part is we don't have the "destination" in `d2`. Will `d2any` read better?

hzhou: `from_dev` is 8 characters. `d2any` is 5 characters. Are we bargaining here? :)

minsii: How about `from_registered_host` and `from_managed`? I am more concerning that it is not consistent with the other subroutines' name.

hzhou: Now you mentioned, were you using `m` to encode `managed`? It was very cryptic to me. The `from_xxx` routines are at one higher level than `_x2y` routines, I think the different name patterns are ok here -- the hierarchy is more important to read than the individuals. As for the two names, how about `from_reghost` (I'll accept `from_rhost` with some reluctance)? `from_managed` seems to be a minimal readable name to me.

hzhou (approved):
My comment on the naming (of 3 functions) remains, but I am not blocking the PR. I leave the decision back to you.

src/backend/src/yaksuri_progress.c
minsii: `from_rhost` seems good to me.

minsii:
> My comment on the naming (of 3 functions) remains, but I am not blocking the PR. I leave the decision back to you.

Sorry I was tied with other work thus cannot follow up since yesterday night. Pushed changes for renaming. Will merge in after all jobs are cleaned.

minsii:
test:yaksa/gpu

PR #181, opened 2021-04-23
backend/cuda: Fix bashism in shell script

### LINKED ISSUE #180: bash syntax in sh script
Hi folks,

https://github.com/pmodels/yaksa/blob/957ff4fc24de2eb87c8892cbd3ec41759c965848/src/backend/cuda/cudalt.sh#L35

The bash substring expansion `${@:2}` is an extension that's not present in all `/bin/sh` implementations.  It caused a `bad substitution` error when I was configuring on a system where `/bin/sh` is dash.

I apologize for not submitting a patch: I didn't try to understand the line that caused the error, I just changed the shebang to `#!/bin/bash` and I thought you wouldn't want to do that for a real fix.

raffenet:
## Pull Request Description

Use shift instead of array slicing syntax. Fixes a build issue on
systems with dash as /bin/sh, e.g. Ubuntu. Closes #180.

## Expected Impact

Fix build issue on Ubuntu.

## Author Checklist

### DIFF
--- a/src/backend/cuda/cudalt.sh
+++ b/src/backend/cuda/cudalt.sh
@@ -14,6 +14,7 @@ fi
 
 LO_FILEPATH="$1"
 O_FILEPATH="${LO_FILEPATH%%.lo}.o"
+shift # handle the rest of the arguments together with ${@}
 
 LO_DIR=$(dirname $O_FILEPATH)
 O_FILENAME=$(basename $O_FILEPATH)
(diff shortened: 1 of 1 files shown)


hzhou:
test:yaksa/gpu

src/backend/cuda/cudalt.sh
```diff
@@ -32,11 +32,12 @@ if test ! -d "$PIC_DIR" ; then
     mkdir -p "$PIC_DIR"
 fi
 
-CMD="${@:2} -Xcompiler -fPIC -o $PIC_FILEPATH"
+shift
```
hzhou, line 35: Could you add a comment on what is being shifted? Maybe it is easier to read if we do the shift right after line 15?

hzhou (approved):
A minor comment. I leave it up to you.

src/backend/cuda/cudalt.sh
raffenet: Sure, can do.

raffenet:
test:yaksa/gpu

PR #184, opened 2021-05-11
frontend/backend: implement blocking pack/unpack and optimize fast-path

minsii:
#### Background
Current API supports only nonblocking `ipack|iunpack` and `request_wait`. The user can use them to implement blocking pack|unpack, however, it causes suboptimal performance in fast-path copy. Performance study with CUDA showed that use of **event-creation-record-destroy** APIs in the current nonblocking approach introduce about **3us~** consistent overhead. Such an overhead is significant at the fast-path of d<->d, d<->m, d<->rh copy which usually takes latency only at 8~15us.

On the user side (e.g., in MPICH), yaksa APIs are often used in a blocking manner depends on the communication protocol or latency/message-rate preference. Thus, using the underlying nonblocking yaksa routines introduce unnecessary overhead. 

#### What is changed
This PR implements the blocking `pack|unpack` API. Thus, yaksa can optimize the blocking use case by skipping event creation. Instead, it uses the `synchronize` backend driver hook. In CUDA, it is `cudaStreamSynchronize`.

## Expected Impact
- For CUDA fast-path, it saves 3+us for blocking pack/unpack use case.
- For CUDA slow-paths, it creates event as of today and the event is blocking waited at end of frontend pack|unpack. Thus, no performance change compared to current version.
- For ZE backend, `synchronize` hook is not defined. Thus, Even fast-path will still create event, same performance as of today.

## Author Checklist

### DIFF
--- a/src/frontend/include/yaksi.h
+++ b/src/frontend/include/yaksi.h
@@ -148,6 +148,7 @@ typedef struct yaksi_type_s {
 typedef struct yaksi_request_s {
     yaksu_handle_t id;
     yaksu_atomic_int cc;        /* completion counter */
+    bool is_blocking;           /* ipack/iunpack are nonblocking; pack/unpack are blocking */
 
     /* give some private space for the backend to store content */
     yaksur_request_s backend;
@@ -264,7 +265,7 @@ int yaksi_type_handle_dealloc(yaksa_type_t handle, yaksi_type_s ** type);
 int yaksi_type_get(yaksa_type_t type, yaksi_type_s ** yaksi_type);
 
 /* request pool */
-int yaksi_request_create(yaksi_request_s ** request);
+int yaksi_request_create(yaksi_request_s ** request, bool is_blocking);
 int yaksi_request_free(yaksi_request_s * request);
 int yaksi_request_get(yaksa_request_t request, yaksi_request_s ** yaksi_request);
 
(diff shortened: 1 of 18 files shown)


src/frontend/include/yaksi.h
```diff
@@ -71,6 +71,11 @@ struct yaksi_request_s;
 #define YAKSI_REQUEST_GET_OBJECT_ID(handle) \
     ((handle) << (64 - YAKSI_REQUEST_OBJECT_ID_BITS) >> (64 - YAKSI_REQUEST_OBJECT_ID_BITS))
 
+typedef enum {
+    YAKSI_REQUEST_NONBLOCKING,
+    YAKSI_REQUEST_BLOCKING,
+} yaksi_request_nb_kind_e;
```
hzhou, line 77: It is enum of boolean values. I found the name yaksi_request_*nb*_kind_e particularly not helping. Why not just use `bool is_blocking` in the interface rather than the indirect enum type?

hzhou (commented):
Looks good. Just one comment for now.

src/frontend/include/yaksi.h
minsii: addressed in the latest push

minsii:
@hzhou I addressed the comment int the latest push. I also updated test/pack/pack.c to enable blocking API tests.

hzhou (approved):
LGTM

PR #189, opened 2021-06-28
type_create: fix checking for contig

hzhou:
## Pull Request Description
The previous code is incorrect in only checking for starting position
between each block in yaksi_type_create_hindexed. We need check whether
previous upper bound is the same as the current lower bound.

Reference: https://github.com/pmodels/mpich/issues/5391

TODO: 
* [x] add tests for checking contig attributes

## Author Checklist

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -24,6 +24,7 @@ error() {
 ########################################################################
 
 genpup_args=
+gentests_args=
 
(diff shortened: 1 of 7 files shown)


hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

yfguo (approved):

PR #192, opened 2021-07-14
Add yaksa_pack_stream and yaksa_unpack_stream

hzhou:
## Pull Request Description

Add `yaksa_pack_stream` and `yaksa_unpack_stream`.

The explicit stream version of routines will only enqueue memory operations and host routines (as stream callbacks).

image

In an optimal case, where the yaksa operation is just a single kernel, we just enqueue that kernel(s). But there are fallback cases that we need enqueue CPU callbacks. For example, we may need run pre/post CPU operations.

## Notes

* The progress routines are rather messy -- duplicating ~1,600 loc. I am focusing on getting the code work, and will leave the refactoring to the future.
* The user-supplied stream has to match the yaksa internal device -- the target device.
* Cuda runtime defines the semantics of stream and stream callback. It is difficult to make an abstraction that works with all potential gpu driver, thus --
    - use `void *` for `stream`. For cuda, that is `cudaStream_t *` -- note this is one additional pointer indirection.  Internally, `cudaStream_t` is already a pointer. We can't directly use `cudaStream_t` because it could be an integer or a `struct`. *bad*: this prevents us to effectively differentiate different streams.
    - use `void (*fn)(void *data)` for host callback function prototype. -- if this is not compatible with other GPU drivers, the backend needs to employ wrapper functions.
* Ideally, we would hope there to be some standardization. For example, define `stream` to be an `integer` and host callbacks be `void (*fn)(void *data)` . Ultimately, I belive at some point an open standardization gonna come along, like `OpenGL`.
* The error recovery is fragile. Since we are queuing up the operations, how would we deal with *unexpected* errors?

## TODO
* [x] Add tests

## Author Checklist

### DIFF
--- a/maint/gentests.py
+++ b/maint/gentests.py
@@ -145,6 +145,8 @@ if __name__ == '__main__':
                        " -num-threads 4")
     gen_pack_iov_tests("pack", "test/pack/testlist.blocking.gen", True, \
                        " -blocking")
+    gen_pack_iov_tests("pack", "test/pack/testlist.stream.gen", True, \
+                       " -stream")
 
     gen_pack_iov_tests("iov", "test/iov/testlist.gen", False)
     gen_pack_iov_tests("iov", "test/iov/testlist.threads.gen", False, " -num-threads 4")
--- a/src/backend/cuda/genpup.py
+++ b/src/backend/cuda/genpup.py
@@ -204,7 +204,7 @@ def generate_host_function(b, darray):
             funcprefix = funcprefix + "%s_" % d
         funcprefix = funcprefix + b.replace(" ", "_")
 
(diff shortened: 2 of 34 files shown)


hzhou:
@yfguo FYI. Still WIP.

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

yfguo (approved):

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou:
@yfguo Could you review the last 6 commits, which are new since your last review? I'll squash all the progress commits after the review.

yfguo (approved):

PR #193, opened 2021-07-22
config: fail if CUDA was requested but cannot be used

spickartz:
## Pull Request Description
In case CUDA was requested explicitly via the `--with-cuda=<PATH>` option, the configure step should fail if it is not functional (i.e., cannot be found or nvcc is not working). Otherwise, third-party tools using Yaksa might run into issues as the resulting build will not include CUDA support although expected.

## Expected Impact
If a wrong path is given to the `--with-cuda` option or the `nvcc` compiler is not functional (e.g., this is currently the case if GCC 11 is used), the configure step will fail.

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -91,6 +91,10 @@ EOF
         fi
         rm -f conftest.*
     fi
+
+    if test "$with_cuda" -a "${have_cuda}" = "no"; then
+        AC_MSG_ERROR([CUDA was requested but it is not functional])
+    fi
 fi
 AM_CONDITIONAL([BUILD_CUDA_BACKEND], [test x${have_cuda} = xyes])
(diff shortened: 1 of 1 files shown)


src/backend/cuda/subconfigure.m4
```diff
@@ -91,6 +91,10 @@ EOF
         fi
         rm -f conftest.*
     fi
+
+    if test "$with_cuda" -a "${have_cuda}" = "no"; then
```
hzhou, line 95:
```suggestion

    if test "$with_cuda" != "" -a "${have_cuda}" = "no"; then
```

hzhou:
@spickartz I made a new PR -- #195, since we need handle the empty `--with-cuda` case as well as the ZE options.

Could you review it https://github.com/pmodels/yaksa/pull/195 ?

hzhou:
Replace by #195

PR #195, opened 2021-07-30
config: fail if CUDA/ZE requested but not found

hzhou:
## Pull Request Description

Enhance the configure library checking so that if cuda or ze were
requested but not found, abort.

## Author Checklist


yfguo:
This should fix the CUDA+ICC problem. NVCC need explicit option, `-ccbin`, to set the host compiler when a non-default compiler is used.

Looking back at previous warning check jobs, the CUDA+ICC configuration was never building the CUDA backend. This should solve the problem.

src/backend/cuda/subconfigure.m4
```diff
@@ -58,17 +58,28 @@ AC_ARG_WITH([cuda-sm],
...
 EOF
-        ${with_cuda}/bin/nvcc -c conftest.cu 2> /dev/null
+        if test -n "$ac_save_CC" ; then
+            NVCC_FLAGS="-ccbin $ac_save_CC"
+            # if the host compiler is PGI, change the -ccbin to pgc++ instead of pgcc
+            NVCC_FLAGS=`echo $NVCC_FLAGS | sed 's/pgcc/pgc++/g'`
+        else
+            NVCC_FLAGS=""
+        fi
+        ${with_cuda}/bin/nvcc $NVCC_FLAGS -c conftest.cu 2> /dev/null
```
spickartz, line 77: Currently, it even finds CUDA via the environment except for the `nvcc` test if `with-cuda` does not contain a path. This would fix it for the test as well:

```suggestion
        # try nvcc from PATH if 'with-cuda' does not contain a valid path
        if test ${with_cuda} == "yes"; then
            nvcc_bin="nvcc"
        else
            nvcc_bin=${with_cuda}/bin/nvcc
        fi

        ${nvcc_bin} $NVCC_FLAGS -c conftest.cu 2> /dev/null
```

yfguo:
@hzhou the tests are clean now. PR is good to go.

src/backend/ze/subconfigure.m4
```diff
@@ -22,7 +22,7 @@ if test "$with_ze" != "no" ; then
     fi
     # ze_api.h relies on support for c11
     if test "${have_ze}" = "yes" ; then
-        PAC_PUSH_FLAG([CFLAGS])
+PAC_PUSH_FLAG([CFLAGS])
```
hzhou, line 25: What is this change?

hzhou (commented):
Looks good. Just one minor comment.

hzhou:
test:yaksa/gpu

src/backend/ze/subconfigure.m4
yfguo: Fixed now.

yfguo:
@hzhou PR cleaned.

hzhou:
Looks good. I can't approve my own PR, but you can @yfguo

yfguo (approved):

PR #196, opened 2021-08-02
info: add info hint mapped_device

hzhou:
## Pull Request Description
For device-to-device operations, the mapped_device hint can be used to
influence which device context (stream) to use.

If the buffer is mapped with ipc, using the stream of mapped_device has
better performance. And in the case of launching kernel, it has to be
launched on the mapped device.

## Author Checklist


yfguo (approved):

hzhou:
test:yaksa/gpu

PR #198, opened 2021-08-24
config: add --enable-fast=all

hzhou:
## Pull Request Description
MPICH uses option --enable-fast=all to enable most common fast options.
However, yaksa ignores it, resulting from worse performance than the default
-O2. Add the option to avoid surprises when building mpich.

Once merged, this should fix the current mpich nightly enable fast test (TIMEOUT in `./rma/large_acc_flush_local`).

## Author Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,6 +146,13 @@ for option in ${enable_fast} ; do
         ndebug)
             PAC_APPEND_FLAG([-DNDEBUG],[CFLAGS])
             ;;
+        all|yes)
+            PAC_APPEND_FLAG([-O2],[CFLAGS])
+            PAC_APPEND_FLAG([-DNDEBUG],[CFLAGS])
+            ;;
+        none|no)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -146,6 +146,13 @@ for option in ${enable_fast} ; do
 	ndebug)
 	    PAC_APPEND_FLAG([-DNDEBUG],[CFLAGS])
 	    ;;
+        all|yes)
+            PAC_APPEND_FLAG([-O2],[CFLAGS])
+	    PAC_APPEND_FLAG([-DNDEBUG],[CFLAGS])
+	    ;;
+        none|no)
+            PAC_APPEND_FLAG([-O0],[CFLAGS])
+	    ;;
```
raffenet, line 155: Change is good, just the whitespace is kinda out of whack.

raffenet (approved):
Approved but please fixup the whitespace before merge.

PR #201, opened 2021-09-04
Optimize backend for contiguous datatype

zhenggb72:
In case of REPLACE operation and contiguous datatype, remove the unnecessary device bounce buffer used for packing or unpacking (which is only needed if it is non-contiguous type). This saves one extra copying for performance.

This PR optimizes for copy between device and unregistered host buffer, and between device to device (w/o p2p).

The third commit fixed a (long standing) bug in backend progress when a packing/unpacking of more than 16MB buffer may exit prematurely (i.e. when the device buffer pool runs out). We have seen validation errors on very large message in various OSU benchmarks due to this bug.

The fourth commit fixed a bug in ze for event_record.

## Author Checklist

### DIFF
--- a/src/backend/src/yaksuri_progress.c
+++ b/src/backend/src/yaksuri_progress.c
@@ -797,31 +797,51 @@ static int pack_d2urh_acquire(yaksuri_request_s * reqpriv, yaksuri_subreq_s * su
 {
     int rc = YAKSA_SUCCESS;
     yaksuri_gpudriver_id_e id = reqpriv->gpudriver_id;
+    yaksa_op_t op = subreq->u.multiple.op;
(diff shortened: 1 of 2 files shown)


hzhou:
test:yaksa/gpu

zhenggb72:
test:yaksa/gpu

zhenggb72:
test:yaksa/gpu

zhenggb72:
test:yaksa/gpu

zhenggb72:
The last commit fixed a bug in backend progress when a packing/unpacking of more than 16MB buffer may exit prematurely (i.e. when buffer pool runs out).

yfguo (approved):

PR #203, opened 2021-09-24
frontend: enable use of -fvisibility

zhenggb72:
Allow --enable-visibility to be used with configure
With -fvisibility=hidden, the entries of dynamic symbol table reduced from 17137 to 35:

```
nm -D ./libyaksa.so |  grep " T " | wc -l
35
```

Hopefully this improves the library loading.
It also reduced the binary by 2.3MB, which however is still huge at 864MB (with ze enabled).

## Author Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -112,6 +112,9 @@ else
     fi
 fi
 
+PAC_CHECK_VISIBILITY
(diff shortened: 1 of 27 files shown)


yfguo (approved):

hzhou:
@zhenggb72 @yfguo This PR will make yaksa symbols public in mpich when embedding yaksa, right?

PR #204, opened 2021-10-04
Disable YAKSA_PUBLIC_API and visibility when in embedded mode

zhenggb72:
Disable visibility attribute when building with embedded mode

## Author Checklist


hzhou:
@zhenggb72 Thanks for the fix! Could you verify -- build with MPICH in embedded mode and verify that yaksa symbols are not in libmpi.so's dynamic symbol table?

zhenggb72:
I tested locally. 
(1) With the original version before even the first visibility PR, this is what in libmpi.so (configure with --enable-visibility):

> readelf -s libmpi.so  | grep ' yaksa_init'
  2517: 0000000000920975 12089 FUNC    GLOBAL DEFAULT   12 yaksa_init
 36677: 0000000000920975 12089 FUNC    GLOBAL DEFAULT   12 yaksa_init

not sure why each function appears twice. But they are all GLOBAL. 
The total number of yaksa API exposed is:

readelf -s libmpi.so | grep FUNC | grep ' yaksa_' | wc
     66     528    5188

it is doubled because each function appears twice.

(2) After applying the previous visibility PR (i.e. adding --enable-visibility and attribute, but ignored the --enable-embedded), everything is the same:

readelf -s libmpi.so | grep FUNC | grep ' yaksa_init'
  2934: 0000000000707ca5 12079 FUNC    GLOBAL DEFAULT   12 yaksa_init
 51700: 0000000000707ca5 12079 FUNC    GLOBAL DEFAULT   12 yaksa_init

readelf -s libmpi.so | grep FUNC | grep ' yaksa' | wc
     66     528    5188


So at least, it does not seem to make it worse.


(3) with this PR:

 readelf -s libmpi.so | grep FUNC | grep ' yaksa_init'
 34060: 0000000000707125 12079 FUNC    LOCAL  DEFAULT   12 yaksa_init

invisibility takes effect now, so it is LOCAL.

 readelf -s libmpi.so | grep FUNC | grep ' yaksa' | wc
     33     264    2594

Number of functions also makes sense now.

PR #207, opened 2021-12-23
util: Do not warn on leaked handles by default

raffenet:
## Pull Request Description

Only warn if handles are leaked in debug mode. MPICH users have noticed
yaksa warnings in default builds, which traditionally have not warned on
leaked MPI handles. This aligns the behavior between the two packages.

## Expected Impact

Silence leak warnings in default builds.

## Author Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -138,6 +138,7 @@ AC_ARG_ENABLE([g],AS_HELP_STRING([--enable-g],[alias for --enable-debug]),,[enab
 AC_ARG_ENABLE([debug],AS_HELP_STRING([--enable-debug],[adds -g to CFLAGS]),,[enable_debug=${enable_g}])
 if test "${enable_debug}" != "no" ; then
     PAC_APPEND_FLAG([-g],[CFLAGS])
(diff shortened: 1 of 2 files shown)


raffenet:
Perhaps this would be fine as a runtime option instead of compile time? Or maybe both? What do you think @hzhou?

hzhou:
> Perhaps this would be fine as a runtime option instead of compile time? Or maybe both? What do you think @hzhou?

Sounds fine. Are you thinking of having an additional flag/hint passing down to yaksa_init? It's some work and no one is requesting it :) So it's up to you

hzhou (approved):
LGTM. As far as mpich is concerned, the `--enable-g` option in MPICH already covers the datatype handle leaks, so it is fine and preferred that yaksa do not warn it again by default.

raffenet:
> LGTM. As far as mpich is concerned, the `--enable-g` option in MPICH already covers the datatype handle leaks, so it is fine and preferred that yaksa do not warn it again by default.

Sounds good. I'll just merge this as-is 😄.

PR #208, opened 2022-01-11
Use Streaming Store For Cross-Socket Memory Copies

wesbland:
## Pull Request Description

This PR is essentially a duplicate of https://github.com/pmodels/mpich/pull/5758, but for data copies in Yaksa rather than in MPICH. I'll copy the description from that PR:

Adds a new type of `memcpy` that uses the AVX instruction `_mm256_stream_si256` to perform a streaming copy that writes data without bringing it into the cache of the writing process. This is used with the iqueue send code to avoid extra latency, especially when writing data from a process located on one socket to a process located on another socket.

## Author Checklist

### DIFF
--- a/src/backend/seq/pup/yaksuri_seq_pup.c
+++ b/src/backend/seq/pup/yaksuri_seq_pup.c
@@ -29,15 +29,17 @@ int yaksuri_seq_ipack(const void *inbuf, void *outbuf, uintptr_t count, yaksi_ty
 {
     int rc = YAKSA_SUCCESS;
     yaksuri_seqi_type_s *seq_type = (yaksuri_seqi_type_s *) type->backend.seq.priv;
+    int memcpy_type = YAKSU_MEMCPY__SYSTEM;
 
     uintptr_t iov_pack_threshold = YAKSURI_SEQI_INFO__DEFAULT_IOV_PUP_THRESHOLD;
     if (info) {
         yaksuri_seqi_info_s *seq_info = (yaksuri_seqi_info_s *) info->backend.seq.priv;
         iov_pack_threshold = seq_info->iov_pack_threshold;
+        if (seq_info->stream_cpy) memcpy_type = YAKSU_MEMCPY__STREAM;
     }
 
     if (op == YAKSA_OP__REPLACE && type->is_contig) {
-        memcpy(outbuf, (const char *) inbuf + type->true_lb, type->size * count);
+        yaksu_memcpy(outbuf, (const char *)inbuf + type->true_lb, type->size * count, memcpy_type);
     } else if (op == YAKSA_OP__REPLACE && type->size / type->num_contig >= iov_pack_threshold) {
         struct iovec iov[MAX_IOV_LENGTH];
         char *dbuf = (char *) outbuf;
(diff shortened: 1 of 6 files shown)


src/backend/seq/pup/yaksuri_seq_pup.c
```diff
@@ -29,15 +29,17 @@ int yaksuri_seq_ipack(const void *inbuf, void *outbuf, uintptr_t count, yaksi_ty
...
     uintptr_t iov_pack_threshold = YAKSURI_SEQI_INFO__DEFAULT_IOV_PUP_THRESHOLD;
     if (info) {
         yaksuri_seqi_info_s *seq_info = (yaksuri_seqi_info_s *) info->backend.seq.priv;
         iov_pack_threshold = seq_info->iov_pack_threshold;
+        if (seq_info->stream_cpy) memcpy_type = YAKSU_MEMCPY__STREAM;
     }
 
     if (op == YAKSA_OP__REPLACE && type->is_contig) {
-        memcpy(outbuf, (const char *) inbuf + type->true_lb, type->size * count);
+        yaksu_memcpy(outbuf, (const char *)inbuf + type->true_lb, type->size * count, memcpy_type);
```
hzhou, line 42: Isn't the contig case bypassed in MPICH already with `MPIR_Memcpy`?

wesbland: Yes. Though it could presumably be done a different way outside of MPICH. This code would work for either case.

hzhou: I am really reluctant to have duplicated code in two places, so if it is unnecessary for now, I would delay it.

wesbland: Do you mean this entire PR or just this line?

hzhou: I think the entire PR is unnecessary for now.

wesbland: It's still useful if you want to accelerate non-contiguous copies. The code in MPICH by itself doesn't handle that situation.

hzhou: Only the low-density datatype is handled in this PR anyway. MPICH can handle that as well if it wants to, and it will probably be simpler than maintaining a duplicate code in yaksa and the extra hint.

wesbland:
Looks like you guys need to update the way Jenkins is hooking up with GitHub. At least some of the tests are failing above because the code can't be fetched.

From https://jenkins-pmrs.cels.anl.gov/job/yaksa-whitespace/622/console:

> fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.

hzhou:
test:yaksa/whitespace

hzhou:
test:yaksa/whitespace

src/backend/seq/pup/yaksuri_seq_pup.c
wesbland: I'm not very familiar with the Yaksa code so I probably missed something here. Where else should I look to handle the remaining cases?

hzhou: They are in the python-generated "pup" functions, in which they are assignments within loops.

wesbland: I see. That's why I didn't find it then. I'll see if I can figure out what needs to change there.

wesbland:
Agree that this is not necessary right now. Though there is an opportunity for improvement in some of the autogenerated code, it's outside of the scope of the first round of changes and I won't have time to do it in the next week. Going to close this.

cc: @zhenggb72

PR #209, opened 2022-01-17
Fix build with Nvidia HPC compilers

raffenet:
## Pull Request Description

Apply same fix we did for building with PGI compilers, since they are the same as nvc/nvc++.

## Expected Impact

Fix CUDA-aware builds with nvc/nvc++.

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -66,9 +66,9 @@ if test "$with_cuda" != "no" ; then
 EOF
         if test -n "$ac_save_CC" ; then
             NVCC_FLAGS="-ccbin $ac_save_CC"
-            # - pgcc doesn't work, use pgc++ instead
+            # - pgcc/nvc doesn't work, use pgc++/nvc++ instead
             # - Extra optins such as `gcc -std=gnu99` doesn't work, strip the option
-            NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
+            NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/nvc/nvc++/g' -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
         else
             NVCC_FLAGS=""
         fi
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -73,10 +73,10 @@ EOF
             NVCC_FLAGS=""
         fi
         # try nvcc from PATH if 'with-cuda' does not contain a valid path
-        if test ${with_cuda} == "yes"; then
-            nvcc_bin="nvcc"
-        else
+        if test -d ${with_cuda} ; then
             nvcc_bin=${with_cuda}/bin/nvcc
+        else
+            nvcc_bin=nvcc
         fi
         ${nvcc_bin} $NVCC_FLAGS -c conftest.cu 2> /dev/null
         if test "$?" = "0" ; then


raffenet:
test:yaksa/warnings

src/backend/cuda/subconfigure.m4
```diff
@@ -66,14 +66,15 @@ if test "$with_cuda" != "no" ; then
...
+            # - pgcc/nvc doesn't work, use pgc++/nvc++ instead
             # - Extra optins such as `gcc -std=gnu99` doesn't work, strip the option
             NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
+            NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/nvc/nvc++/g' -e's/ -std=.*//g')
         else
             NVCC_FLAGS=""
         fi
         # try nvcc from PATH if 'with-cuda' does not contain a valid path
-        if test ${with_cuda} == "yes"; then
+        if test ${with_cuda} = "yes"; then
```
hzhou, line 77: Wait, shouldn't this be `!=`?

src/backend/cuda/subconfigure.m4
```diff
@@ -66,14 +66,15 @@ if test "$with_cuda" != "no" ; then
 EOF
         if test -n "$ac_save_CC" ; then
             NVCC_FLAGS="-ccbin $ac_save_CC"
-            # - pgcc doesn't work, use pgc++ instead
+            # - pgcc/nvc doesn't work, use pgc++/nvc++ instead
             # - Extra optins such as `gcc -std=gnu99` doesn't work, strip the option
             NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
+            NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/nvc/nvc++/g' -e's/ -std=.*//g')
```
hzhou, line 72: You may combine the two cases with `sed -e 's/\(pgcc\|nvc\)/\1++/g'`, Or simply add another `-e 's/nvc/nvc++/g' to `sed`.

src/backend/cuda/subconfigure.m4
raffenet: Lol, yeah it should looks like it.

raffenet: Oh wait, I see it now. `yes` isn't a valid path so it just tries to use whatever it can find in `PATH`.

raffenet: Maybe it would be better to do:
```
if test -d ${with_cuda} ; then
    nvcc_bin=${with_cuda}/bin/nvcc
else
    nvcc_bin=nvcc
fi
```

src/backend/cuda/subconfigure.m4
raffenet: done

src/backend/cuda/subconfigure.m4
hzhou: Yes, that reads much clearer.

hzhou: Or maybe it is more proper to use `AC_CHECK_PROG`?

hzhou (approved):
Still like to make the nvcc check more readable, but it's minor and I leave it to you

src/backend/cuda/subconfigure.m4
raffenet: I went with the `test -d` version. AC_CHECK_PROG seems a little clunky if I am understanding it right.

PR #210, opened 2022-01-19
config: Improvements to nvcc test

raffenet:
## Pull Request Description

Add a way to specify user flags for nvcc. Improve logging of compilation test.

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -81,7 +81,7 @@ if test "$with_cuda" != "no" ; then
         AC_COMPILE_IFELSE([AC_LANG_PROGRAM([__global__ void foo(int x) {}],[])],
         [
             AC_DEFINE([HAVE_CUDA],[1],[Define is CUDA is available])
-            AS_IF([test -n "${with_cuda}"],[NVCC=${with_cuda}/bin/nvcc],[NVCC=nvcc])
+            AS_IF([test -d "${with_cuda}"],[NVCC=${with_cuda}/bin/nvcc],[NVCC=nvcc])
             AC_SUBST(NVCC)
             AC_SUBST(NVCC_FLAGS)
             AC_MSG_RESULT([yes])
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -82,28 +82,30 @@ EOF
             AS_IF([test -n "${with_cuda}"],[NVCC=${with_cuda}/bin/nvcc],[NVCC=nvcc])
             AC_SUBST(NVCC)
             AC_SUBST(NVCC_FLAGS)
-            # nvcc compiled applications need libstdc++ to be able to link
-            # with a C compiler
-            PAC_PUSH_FLAG([LIBS])
-            PAC_APPEND_FLAG([-lstdc++],[LIBS])
-            AC_LINK_IFELSE(
-                [AC_LANG_PROGRAM([int x = 5;],[x++;])],
-                [libstdcpp_works=yes],
-                [libstdcpp_works=no])
-            PAC_POP_FLAG([LIBS])
-            if test "${libstdcpp_works}" = "yes" ; then
-                PAC_APPEND_FLAG([-lstdc++],[LIBS])
-                AC_MSG_RESULT([yes])
-            else
-                have_cuda=no
-                AC_MSG_RESULT([no])
-            fi
+            AC_MSG_RESULT([yes])
         else
             have_cuda=no
             AC_MSG_RESULT([no])
             AC_MSG_ERROR([CUDA was not functional with provided host compiler (CC)])
         fi
         rm -f conftest.*
+        # nvcc compiled applications need libstdc++ to be able to link
+        # with a C compiler
+        AC_MSG_CHECKING([if $CC can link libstdc++])
+        PAC_PUSH_FLAG([LIBS])
+        PAC_APPEND_FLAG([-lstdc++],[LIBS])
+        AC_LINK_IFELSE(
+            [AC_LANG_PROGRAM([int x = 5;],[x++;])],
+            [libstdcpp_works=yes],
+            [libstdcpp_works=no])
+        PAC_POP_FLAG([LIBS])
+        if test "${libstdcpp_works}" = "yes" ; then
+            PAC_APPEND_FLAG([-lstdc++],[LIBS])
+            AC_MSG_RESULT([yes])
+        else
+            have_cuda=no
+            AC_MSG_RESULT([no])
+        fi
     fi
     if test "${have_cuda}" = "no" -a "$with_cuda" != ""; then
         AC_MSG_ERROR([CUDA was requested but it is not functional])
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -61,9 +61,6 @@ if test "$with_cuda" != "no" ; then
     PAC_CHECK_HEADER_LIB([cuda_runtime_api.h],[cudart],[cudaStreamSynchronize],[have_cuda=yes],[have_cuda=no])
     if test "${have_cuda}" = "yes" ; then
         AC_MSG_CHECKING([whether nvcc works])
-        cat>conftest.cu<<EOF
-        __global__ void foo(int x) {}
-EOF
         if test -n "$ac_save_CC" ; then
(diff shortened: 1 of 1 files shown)


raffenet:
Example of improved output in `config.log`. Now you can see what command was executed.
```
configure:17384: checking whether nvcc works
configure:17417: /nfs/gce/software/custom/linux-centos7-x86_64/hpc_sdk/21.11/Linux_x86_64/21.11/cuda/bin/nvcc -ccbin pgc++ -c conftest.cu >&5
configure:17417: $? = 0
```

raffenet:
And an example from an unsuccessful test
```
configure:14407: checking whether nvcc works
configure:14440: /opt/nvidia/hpc_sdk/Linux_x86_64/22.1/cuda/11.5/bin/nvcc -ccbin nvc++ -c conftest.cu >&5
"/opt/nvidia/hpc_sdk/Linux_x86_64/22.1/cuda/11.5/bin/../targets/x86_64-linux/include/crt/host_config.h", line 118: catastrophic error: #error directive: -- unsupported pgc++ configuration! Only pgc++ 18, 19, 20 and 21 are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
  #error -- unsupported pgc++ configuration! Only pgc++ 18, 19, 20 and 21 are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
   ^

1 catastrophic error detected in the compilation of "conftest.cu".
Compilation terminated.
configure:14440: $? = 2
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "yaksa"
| #define PACKAGE_TARNAME "yaksa"
| #define PACKAGE_VERSION "unreleased"
| #define PACKAGE_STRING "yaksa unreleased"
| #define PACKAGE_BUGREPORT ""

<snip>

| /* end confdefs.h.  */
| __global__ void foo(int x) {}
| int
| main ()
| {
|
|   ;
|   return 0;
| }
configure:14522: result: no
configure:14524: error: CUDA was not functional with provided host compiler (CC)
```

src/backend/cuda/subconfigure.m4
```diff
@@ -61,25 +61,24 @@ if test "$with_cuda" != "no" ; then
...
             nvcc_bin=nvcc
         fi
-        ${nvcc_bin} $NVCC_FLAGS -c conftest.cu 2> /dev/null
-        if test "$?" = "0" ; then
+        AC_LANG_PUSH([C])
+        ac_ext=cu
+        ac_compile='$nvcc_bin $NVCC_FLAGS -c conftest.$ac_ext >&5'
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([__global__ void foo(int x) {}],[])],
+        [
+            AC_LANG_POP([C])
```
hzhou, line 81: Unbalanced PUSH/POP. The POP will be missed in the failure case, right?

hzhou (commented):
The improvement is so much better!

Just a minor issue.

src/backend/cuda/subconfigure.m4
raffenet: Yes, but the failure case calls `AC_MSG_ERROR`.

raffenet: I can put the AC_LANG_POP([C]) in the failure case anyway. It won't do any harm.

hzhou: Why not move AC_LANG_POP outside `AC_COMPILE_IFELSE`?

hzhou: Oh, I see. Are we doing CXX here?

raffenet: That was my original thought, but in the success case we then call AC_LINK_IFELSE with a C program.

raffenet: Maybe that portion should be moved out, too. Let me take one more crack.

hzhou: I guess I don't understand why we need AC_LANG_PUSH([C]) in the first place. We only configured for C in yaksa, right?

raffenet: My idea was to save the current language settings, then customize `ac_compile` and `ac_ext` to work with CUDA since there is no builtin autoconf support for it. Then AC_LANG_POP to restore the previous settings. Perhaps there is a better way?

hzhou: Now I see. Definitely need a comment here to help the understanding. But I do agree that I should have seen that. :) 

How about move the libstdcpp test outside and after the AC_COMPILE_IFELSE test. Since the other branch is just AC_MSG_ERROR, we can avoid the nesting and simplify the script a bit, right?

raffenet: Will do. Noticed one more issue I'll address in a 3rd commit, too.

raffenet: Now 4 commits, since I split out the un-nesting part from the AC_COMPILE_IFELSE usage.

src/backend/cuda/subconfigure.m4
```diff
@@ -61,51 +61,55 @@ if test "$with_cuda" != "no" ; then
...
+        AC_LANG_PUSH([C])
+        ac_ext=cu
+        ac_compile='$nvcc_bin $NVCC_FLAGS -c conftest.$ac_ext >&5'
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([__global__ void foo(int x) {}],[])],
+        [
             AC_DEFINE([HAVE_CUDA],[1],[Define is CUDA is available])
-            AS_IF([test -n "${with_cuda}"],[NVCC=${with_cuda}/bin/nvcc],[NVCC=nvcc])
+            AS_IF([test -d "${with_cuda}"],[NVCC=${with_cuda}/bin/nvcc],[NVCC=nvcc])
             AC_SUBST(NVCC)
             AC_SUBST(NVCC_FLAGS)
```
hzhou, line 86: Both AC_SUBST probably should move out since they are not part of the IFELSE logic.

src/backend/cuda/subconfigure.m4
```diff
@@ -61,51 +61,55 @@ if test "$with_cuda" != "no" ; then
...
+            have_cuda=no
+            AC_MSG_RESULT([no])
+            AC_MSG_ERROR([CUDA was not functional with provided host compiler (CC)])
+        ])
+        # done with CUDA, back to C
+        AC_LANG_POP([C])
+
+        # nvcc compiled applications need libstdc++ to be able to link
+        # with a C compiler
+        AC_MSG_CHECKING([if we can link libstdc++])
```
hzhou, line 98:
```suggestion
        AC_MSG_CHECKING([if nvcc can link libstdc++])
```

hzhou (approved):
Looks good. Two more nitpicks. Approved and leave the PR to you.

src/backend/cuda/subconfigure.m4
raffenet: I'll save this one for another PR. I notice that yaksa needs an update to silence all the autoconf deprecated warnings. Will handle together with that.

src/backend/cuda/subconfigure.m4
raffenet: I guess the more correct thing is `if $CC can link libstdc++`? I believe what yaksa does is link together CUDA and C object files using the C compiler with `-lstdc++`. Maybe we should actually be linking with a C++ compiler? That addition seemed fishy to me at the time, but it seems to work 🤷.

hzhou: Ahh, that makes sense.

raffenet: I suspect from quick google that we might want to switch to linking with `nvcc`. Libtool appears to have support for it.

hzhou: I am more leaning toward linking with an actual nvcc compiled object. We need test what we are going to do.

raffenet: Yeah, I think what I am getting at is we shouldn't need to manually add libstdc++. From what I can tell, CUDA programming guides don't suggest it. My guess is as long as we link with `nvcc`, the right dependencies will be pulled in. The C compiler doesn't know to bring in any C++ libraries at link time.

Issue #212, opened 2022-01-27
cuda: delay resource creation


raffenet:
See MPICH issue: https://github.com/pmodels/mpich/issues/5810

I am thinking if we can just replace direct accesses to the global cuda stream array
```
$ git grep yaksuri_cudai_global.stream
hooks/yaksuri_cuda_init_hooks.c:        cerr = cudaStreamDestroy(yaksuri_cudai_global.stream[i]);
hooks/yaksuri_cuda_init_hooks.c:    free(yaksuri_cudai_global.stream);
hooks/yaksuri_cuda_init_hooks.c:    yaksuri_cudai_global.stream = (cudaStream_t *)
hooks/yaksuri_cuda_init_hooks.c:        /* cerr = cudaStreamCreateWithFlags(&yaksuri_cudai_global.stream[i], cudaStreamNonBlocking); */
pup/yaksuri_cudai_event.c:    cerr = cudaEventRecord(event->event, yaksuri_cudai_global.stream[device]);
pup/yaksuri_cudai_event.c:    cerr = cudaEventRecord(event, yaksuri_cudai_global.stream[device1]);
pup/yaksuri_cudai_event.c:    cerr = cudaStreamWaitEvent(yaksuri_cudai_global.stream[device2], event, 0);
pup/yaksuri_cudai_pup.c:                                           &yaksuri_cudai_global.stream[target]);
pup/yaksuri_cudai_pup.c:                                             &yaksuri_cudai_global.stream[target]);
pup/yaksuri_cudai_pup.c:    cerr = cudaStreamSynchronize(yaksuri_cudai_global.stream[target]);
```

 with an inline function like
```
cudaStream_t yaksuri_cudai_get_stream(int device);
```

The CUDA docs state the the default stream is 0 or NULL. Can I calloc the stream array and do this to create the stream when needed?
```c
static inline cudaStream_t yaksuri_cudai_get_stream(int device) {
    if (yaksuri_cudai_global.stream[device] == 0) {
        int cur_device;
        cudaGetDevice(&cur_device);

        cudaSetDevice(device);
        cudaStreamCreateWithFlags(&yaksuri_cudai_global.stream[device], cudaStreamNonBlocking);

        cudaSetDevice(cur_device);
    }
    return yaksuri_cudai_global.stream[device];
}
```

raffenet:
[...]

PR #214, opened 2022-02-24
cuda: Delay stream creation

raffenet:
## Pull Request Description

Some applications may seutp GPU affinity after MPI initialization. Delay
CUDA stream creation during yaksa init to avoid allocating resources on
GPUs a process may never use. See #212.

Signed-off-by: Ken Raffenetti <raffenet@mcs.anl.gov>

## Expected Impact

Avoid allocating resources on devices that will not be used.

## Author Checklist

### DIFF
--- a/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
+++ b/src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
@@ -65,17 +65,17 @@ static int finalize_hook(void)
     cudaError_t cerr;
 
     for (int i = 0; i < yaksuri_cudai_global.ndevices; i++) {
-        if (yaksuri_cudai_global.stream[i] != 0) {
+        if (yaksuri_cudai_global.streams[i].created) {
             cerr = cudaSetDevice(i);
             YAKSURI_CUDAI_CUDA_ERR_CHKANDJUMP(cerr, rc, fn_fail);
 
-            cerr = cudaStreamDestroy(yaksuri_cudai_global.stream[i]);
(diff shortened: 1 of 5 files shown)


raffenet:
test:yaksa/gpu

hzhou:
test:yaksa/gpu

src/backend/cuda/hooks/yaksuri_cuda_init_hooks.c
```diff
@@ -141,8 +141,7 @@ int yaksuri_cuda_init_hook(yaksur_gpudriver_hooks_s ** hooks)
         }
     }
 
-    yaksuri_cudai_global.stream = (cudaStream_t *)
-        calloc(yaksuri_cudai_global.ndevices, sizeof(cudaStream_t));
+    yaksuri_cudai_global.streams = calloc(yaksuri_cudai_global.ndevices, sizeof(cudai_stream));
```
raffenet, line 144: we can convert this to regular `malloc` now since we track the state elsewhere

hzhou: `cudai_stream` is an internal struct (pair) that includes the boolean flag. If we use malloc, then we need init the flags to 0 separately.

raffenet: Oh, right. Forgot the boolean was embedded along with the stream handle. Carry on.

hzhou (approved):
LGTM

PR #224, opened 2022-05-12
genpup: fix code generation for pup_max_nesting < 2

yfguo:
## Pull Request Description

Current codes fail to build when `--pup-max-nesting` is set to less than 2 due to incorrect/inconsistent behavior in genpup code:
1. `gencomm.populate_pupfns` always generates as max nesting level == 2
2. `genpup.py` always generates two-level-nesting kernels
3. `Makefile.pup.mk` always generates with targets of two-level-nesting kernels
4. `yaksuri_*i_pup.h` is generated correctly

Generating code with max nesting level < 2 and building will results in
1. compilation error in pupfns code because the missing decls for two-level-nesting kernels in `yaksuri_*i_pup.h`
2. wasted compilation for two-level-nesting kernels that should never be used.

This PR fixes this by adding necessary checks to the code generation. 

## Author Checklist

### DIFF
--- a/src/backend/gencomm.py
+++ b/src/backend/gencomm.py
@@ -161,23 +161,25 @@ def populate_pupfns(pup_max_nesting, backend, blklens, builtin_types, builtin_ma
         yutils.display(OUTFILE, "\n")
 
     yutils.display(OUTFILE, "switch (type->kind) {\n")
-    for dtype1 in derived_types:
-        yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype1.upper())
-        yutils.display(OUTFILE, "switch (type->u.%s.child->kind) {\n" % dtype1)
-        for dtype2 in derived_types:
-            yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype2.upper())
-            yutils.display(OUTFILE, "rc = yaksuri_%si_populate_pupfns_%s_%s(type);\n" % (backend, dtype1, dtype2))
+    if pup_max_nesting > 0:
+        for dtype1 in derived_types:
+            yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype1.upper())
+            yutils.display(OUTFILE, "switch (type->u.%s.child->kind) {\n" % dtype1)
+            if pup_max_nesting > 1:
+                for dtype2 in derived_types:
+                    yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype2.upper())
+                    yutils.display(OUTFILE, "rc = yaksuri_%si_populate_pupfns_%s_%s(type);\n" % (backend, dtype1, dtype2))
+                    yutils.display(OUTFILE, "break;\n")
+                    yutils.display(OUTFILE, "\n")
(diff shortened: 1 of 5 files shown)


src/backend/gencomm.py
```diff
@@ -161,23 +161,25 @@ def populate_pupfns(pup_max_nesting, backend, blklens, builtin_types, builtin_ma
         yutils.display(OUTFILE, "\n")
 
     yutils.display(OUTFILE, "switch (type->kind) {\n")
-    for dtype1 in derived_types:
-        yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype1.upper())
-        yutils.display(OUTFILE, "switch (type->u.%s.child->kind) {\n" % dtype1)
-        for dtype2 in derived_types:
-            yutils.display(OUTFILE, "case YAKSI_TYPE_KIND__%s:\n" % dtype2.upper())
-            yutils.display(OUTFILE, "rc = yaksuri_%si_populate_pupfns_%s_%s(type);\n" % (backend, dtype1, dtype2))
+    if pup_max_nesting > 0:
```
raffenet, line 164: Do we really want to support nesting = 0?

yfguo: Oh, I probably should manually test the code with ZE today. I haven't done it last night.

raffenet: > Nesting = 0 can be useful for dev since it only handles builtin types in backend and have frontend to handle all derived dtypes. So it builts really fast for dev purpose.
> 
> I would say we don't need to unsupport it.

OK, thanks. I didn't realize it still had some functionality.

yfguo: @raffenet Works fine with ZE, expected failures as the main branch.

raffenet (approved):

src/backend/gencomm.py
```diff
@@ -313,10 +319,12 @@ def populate_pupfns(pup_max_nesting, backend, blklens, builtin_types, builtin_ma
     yutils.display(OUTFILE, "#ifndef YAKSURI_%sI_POPULATE_PUPFNS_H_INCLUDED\n" % backend.upper())
     yutils.display(OUTFILE, "#define YAKSURI_%sI_POPULATE_PUPFNS_H_INCLUDED\n" % backend.upper())
     yutils.display(OUTFILE, "\n")
-    for dtype1 in derived_types:
-        for dtype2 in derived_types:
-            yutils.display(OUTFILE, "int yaksuri_%si_populate_pupfns_%s_%s(yaksi_type_s * type);\n" % (backend, dtype1, dtype2))
-        yutils.display(OUTFILE, "int yaksuri_%si_populate_pupfns_%s_builtin(yaksi_type_s * type);\n" % (backend, dtype1))
+    if pup_max_nesting > 0:
+        for dtype1 in derived_types:
+            if pup_max_nesting > 1:
```
hzhou, line 324: This makes depth=1 equivalent to depth=0, which can be simplified by cosolidate the condition to outer layer.

How about rename original `populate_pupfns` to `populate_pupfns_level_2` and add `populate_pupfns_level_0` and `populate_pupfns_level_1` and have `populate_pupfns` conditionally call each? `populate_pupfns_level_0` is likely to be empty and can be omitted.

hzhou: I see the depth=1 is being generated. I guess the patch is fine. Sorry for the noise.

PR #231, opened 2022-09-21
hip: improve --with-hip-sm option 

yfguo:
## Pull Request Description

Adding option to let HIPCC auto-detect target id. Cleaned up option help strings.

Tested and confirmed fixing building problem on Crusher.

#222 would not be needed once this comes in.

## Author Checklist

### DIFF
--- a/src/backend/hip/subconfigure.m4
+++ b/src/backend/hip/subconfigure.m4
@@ -11,38 +11,12 @@
 # --with-hip-sm
 AC_ARG_WITH([hip-sm],
             [
-  --with-hip-sm=<options> (https://reviews.llvm.org/differential/changeset/?ref=2126544)
+  --with-hip-sm=<options>
           Comma-separated list of below options:
-                all - build compatibility for all GPUs supported by the HIP version (can increase compilation time)
-
-                # Kepler architecture
-                AMDGCN - build compatibility for all AMD GPUs
-                gfx600 - GFX600-DAG
-                gfx601 - GFX601-DAG
-                gfx700 - GFX700-DAG
-                gfx701 - GFX701-DAG
-                gfx702 - GFX702-DAG
-                gfx703 - GFX703-DAG
-                gfx704 - GFX704-DAG
-                gfx801 - GFX801-DAG
-                gfx802 - GFX802-DAG
-                gfx803 - GFX803-DAG
-                gfx810 - GFX810-DAG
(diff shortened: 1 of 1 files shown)


yfguo:
Didn't notice there is #222. Will check that out.

src/backend/hip/subconfigure.m4
```diff
@@ -11,38 +11,23 @@
 # --with-hip-sm
 AC_ARG_WITH([hip-sm],
             [
-  --with-hip-sm=<options> (https://reviews.llvm.org/differential/changeset/?ref=2126544)
+  --with-hip-sm=<options>
           Comma-separated list of below options:
+                auto - letting HIPCC to auto detect target_id
```
raffenet, line 16: auto-detection is a great additional. the "all" target took forever to build in my experience.
```suggestion
                auto - let HIPCC auto detect the target_id
```

hzhou: Yanfei and I had discussed offline for a bit. I think we can simply remove the "all" and the subsets options. By default, it is essentially "auto", where `hipcc` do whatever its default. Then user always can pass in specific list of sm, which we directly pass on to `hipcc`. I don't think it is worth while for us to maintain a separate option behavior from `hipcc`'s.

src/backend/hip/subconfigure.m4
```diff
@@ -121,25 +95,28 @@ if test "${have_hip}" = "yes" ; then
...
+                HIP_GENCODE="--offload-arch=${sm}"
+            else
+                HIP_GENCODE="${HIP_GENCODE} --offload-arch=${sm}"
+            fi
+        done
         if test -z "${HIP_GENCODE}" ; then
-            HIP_GENCODE="--offload-arch=${sm}"
-        else
-            HIP_GENCODE="${HIP_GENCODE} --offload-arch=${sm}"
+            AC_MSG_ERROR([--with-hip-sm not specified; either specify it or disable hip support])
```
hzhou, line 116: I am thinking that the default is `auto`.  If the user do `--with-hip-sm=xxx`, we directly pass to `--offload-arch=xxx`.

yfguo: Yeah, the message here is obsolete. We receive the option as comma-separated list, but HIPCC takes them as separate `--offload-arch=` options.

hzhou (approved):
LGTM

PR #234, opened 2022-10-24
Support mixed memory types

spickartz:
## Pull Request Description

Currently, yaksa assumes all memory regions of derived datatypes to have the same memory type. However, in cases with absolute addressing, this is not necessarily true and therefore the caching of the memory type should be disabled. These patches avoid the caching of memory type information in such cases. Additionally, to support different memory types when subrequests are used, the `gpudriver_id` is stored on this level.

See also: https://github.com/pmodels/mpich/pull/6226

## Expected Impact
In case of absolute addressing, these patches will result in an increased amount of memory type queries.

## Author Checklist

### DIFF
--- a/src/backend/src/yaksur_pup.c
+++ b/src/backend/src/yaksur_pup.c
@@ -10,12 +10,13 @@
 #include "yaksuri.h"
 
 static int ipup(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s * type,
-                yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request)
+                yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request,
+                bool always_query_ptr_attr)
 {
     int rc = YAKSA_SUCCESS;
     yaksuri_request_s *reqpriv = (yaksuri_request_s *) request->backend.priv;
 
-    if (reqpriv->gpudriver_id != YAKSURI_GPUDRIVER_ID__UNSET)
+    if (!always_query_ptr_attr && reqpriv->gpudriver_id != YAKSURI_GPUDRIVER_ID__UNSET)
         goto query_done;
 
     yaksuri_info_s *infopriv;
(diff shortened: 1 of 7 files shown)


src/backend/src/yaksur_pup.c
```diff
@@ -10,12 +10,13 @@
...
 static int ipup(const void *inbuf, void *outbuf, uintptr_t count, yaksi_type_s * type,
-                yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request)
+                yaksi_info_s * info, yaksa_op_t op, yaksi_request_s * request,
+                bool cache_gpudriver_id)
 {
     int rc = YAKSA_SUCCESS;
     yaksuri_request_s *reqpriv = (yaksuri_request_s *) request->backend.priv;
 
-    if (reqpriv->gpudriver_id != YAKSURI_GPUDRIVER_ID__UNSET)
+    if (cache_gpudriver_id && reqpriv->gpudriver_id != YAKSURI_GPUDRIVER_ID__UNSET)
```
hzhou, line 19: `cache_gpudirver_id` seems to be a bit misleading, since `reqpriv->gpudriver_id` is still set and used. The direct intention is to "always_query_attr", right?

src/backend/src/yaksuri_progress.c
```diff
@@ -2963,6 +2963,7 @@ int yaksuri_progress_enqueue(const void *inbuf, void *outbuf, uintptr_t count, y
     if (!subreq) {
         goto fn_exit;
     }
+    subreq->gpudriver_id = id;
```
hzhou, line 2966: Could you add a comment that "reqpriv->gpudriver_id" may be different between multiple `yaksuri_progress_enqueue` ?

src/backend/src/yaksur_pup.c
spickartz: Yes, you are right. I changed this to `always_query_ptr_attr`.

src/backend/src/yaksuri_progress.c
spickartz: I added an according comment.

hzhou:
LGTM. Run a gpu test to confirm

test:yaksa/gpu

hzhou:
test:yaksa/gpu

hzhou (approved):
LGTM

PR #240, opened 2023-01-09
cuda: allow user to set NVCC explicitly

spickartz:
## Pull Request Description
We cannot use icc as the host compiler since it conflicts with recent CUDA versions. Therefore, we rely on the default GCC as it is binary compatible with icc.

This is done by simply omitting to set the `-ccbin icc`.  The test in `subconfigure.m4` can also be successfully passed by adding a `-std=c++11`.  However, this still prevents yaksa from being build by using icc as the host compiler.

## Expected Impact
CUDA code will be compiled by using GCC as the host compiler if icc is used within the build. 

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -87,7 +87,7 @@ if test "$with_cuda" != "no" ; then
         # save language settings, customize ac_ext and ac_compile to support CUDA
         AC_LANG_PUSH([C])
         ac_ext=cu
-        ac_compile='$NVCC -c conftest.$ac_ext >&5'
+        ac_compile="$NVCC -c conftest.$ac_ext >&5"
         AC_MSG_CHECKING([whether nvcc works])
         AC_COMPILE_IFELSE([AC_LANG_PROGRAM([__global__ void foo(int x) {}],[])],
         [
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -57,28 +57,38 @@ AC_ARG_WITH([cuda-sm],
 
 
 # --with-cuda
+AC_ARG_VAR([NVCC], [nvcc compiler to use])
 PAC_SET_HEADER_LIB_PATH([cuda])
 if test "$with_cuda" != "no" ; then
     PAC_CHECK_HEADER_LIB([cuda_runtime_api.h],[cudart],[cudaStreamSynchronize],[have_cuda=yes],[have_cuda=no])
     if test "${have_cuda}" = "yes" ; then
-        AC_MSG_CHECKING([whether nvcc works])
-        if test -n "$ac_save_CC" ; then
-            NVCC_FLAGS="$NVCC_FLAGS -ccbin $ac_save_CC"
-            # - pgcc/nvc doesn't work, use pgc++/nvc++ instead
-            # - Extra optins such as `gcc -std=gnu99` doesn't work, strip the option
-            NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/nvc/nvc++/g' -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
-        fi
-        # try nvcc from PATH if 'with-cuda' does not contain a valid path
-        if test -d ${with_cuda} ; then
-            nvcc_bin=${with_cuda}/bin/nvcc
+        # check if NVCC has been set by the user
+        if test -z "$NVCC" ; then
+            if test -n "$ac_save_CC" ; then
+                NVCC_FLAGS="$NVCC_FLAGS -ccbin $ac_save_CC"
+                # - pgcc/nvc doesn't work, use pgc++/nvc++ instead
+                # - Extra optins such as `gcc -std=gnu99` doesn't work, strip the option
+                NVCC_FLAGS=$(echo $NVCC_FLAGS | sed -e 's/nvc/nvc++/g' -e 's/pgcc/pgc++/g' -e's/ -std=.*//g')
+            fi
+
+            # try nvcc from PATH if 'with-cuda' does not contain a valid path
+            if test -d ${with_cuda} ; then
+                nvcc_bin=${with_cuda}/bin/nvcc
+            else
+                nvcc_bin=nvcc
+            fi
+
+            # append the flags to be passed to nvcc
+            NVCC="$nvcc_bin $NVCC_FLAGS"
         else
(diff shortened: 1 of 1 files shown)


src/backend/cuda/subconfigure.m4
```diff
@@ -62,7 +62,9 @@ if test "$with_cuda" != "no" ; then
     PAC_CHECK_HEADER_LIB([cuda_runtime_api.h],[cudart],[cudaStreamSynchronize],[have_cuda=yes],[have_cuda=no])
     if test "${have_cuda}" = "yes" ; then
         AC_MSG_CHECKING([whether nvcc works])
-        if test -n "$ac_save_CC" ; then
+        # only explicitly set the host compiler if not using the icc Intel
+        # compiler (this conflicts with current versions of nvcc)
+        if test -n "$ac_save_CC" -a "${ac_save_CC}" != "icc"; then
```
hzhou, line 67: Did you try `icpc` -- similar to what did for `pgcc` and `nvc`?

spickartz: No, this does not work either. You can simply reproduce this by running:
```console
$ echo '__global__ void foo(int x) {}' | nvcc -x cu -c -ccbin icpc -
```

Instead the more recent `icx` and `icpx` compilers are working (given `-allow-unsupported-compiler` is set; which is probably something we do not want).

spickartz:
I changed the logic to only try the fallback to the default host compiler if it fails with the specified `-ccbin` in the first place. This way `icc` would still be used if the version being tested is compatible with the installed CUDA version. What do you think @hzhou?

src/backend/cuda/subconfigure.m4
```diff
@@ -90,6 +92,29 @@ if test "$with_cuda" != "no" ; then
             have_cuda=no
```
hzhou, line 92: How about move the retest here?

spickartz: Did you mean it this way? I now additionally set the host compiler explicitly to gcc instead of relying on the default.

src/backend/cuda/subconfigure.m4
```diff
@@ -89,7 +91,30 @@ if test "$with_cuda" != "no" ; then
         ],[
             have_cuda=no
             AC_MSG_RESULT([no])
+
+            # if using the Intel icc compiler retry without the '-ccbin' option and fallback to
+            # binary compatible gcc
+            if test "${ac_save_CC}" == "icc"; then
+                NVCC_FLAGS="$save_NVCC_FLAGS -ccbin gcc"
```
hzhou, line 98: I am thinking that we can just try the default (remove the `-ccbin` option) and see if it works. I would rather not to maintain the extra logic of checking `icc` and picking `gcc`.

spickartz: Okay. I was just thinking that we can ensure binary compatibility  this way (with the C code of the non-CUDA files compiled using `icc`). Or don't you see an issue there?

hzhou: How about when the check failed here we simply spit a "helpful" message and ask users to set `NVCC` and re-configure? We did this in `test/mpi/configure` but neglected to do the same for yaksa -- https://github.com/pmodels/mpich/pull/5968

spickartz: Mh aren't we having the exact the same issue [here](https://github.com/pmodels/mpich/blob/main/test/mpi/configure.ac#L736) and even omit the test during configure time? I would assume (did not try it out yet) that the compile would fail when `icc` is used as host compiler.

Probably it is a matter of taste. Do we wann to provide a fallback (and additionally print a warning in such cases). Or do we rather want the user to explicitly set the host compiler if the currently chosen compiler conflicts with `nvcc`?

hzhou: I think we should always have a mechanism to allow users to apply their own settings. So only do the check when `NVCC` is not set by the user. Since users can fallback on that, I think we can just fail for those corner cases such as `icc`.

Yes, we should test the `ccbin` at least once in configure.

spickartz: Okay, I changed it accordingly. Users may now explicitly  set `NVCC` (including flags/options). The retry has been removed.

spickartz: What do you think? Is this was you intended?

src/backend/cuda/subconfigure.m4
```diff
@@ -57,28 +57,36 @@ AC_ARG_WITH([cuda-sm],
...
         else
-            nvcc_bin=nvcc
+            nvcc_bin=$NVCC
+            AC_MSG_WARN([Using user-provided nvcc: '${nvcc_bin}'])
         fi
 
         # save language settings, customize ac_ext and ac_compile to support CUDA
         AC_LANG_PUSH([C])
         ac_ext=cu
         ac_compile='$nvcc_bin $NVCC_FLAGS -c conftest.$ac_ext >&5'
```
hzhou, line 88: Sorry for the late response. How about make it `ac_compile="$NVCC -c ...`, i.e. `NVCC="$nvcc_bin $NVCC_FLAGS"`? The intention is to allow user fully overwrite `NVCC`.

I am surprised to see single quotes here. Wouldn't the single quote prevent shell variable expansion? What did I miss?

hzhou: > > I am surprised to see single quotes here. Wouldn't the single quote prevent shell variable expansion? What did I miss?
> 
> Mh good point, oversaw that but it is working as expected, e.g., I saw something like `configure:18564: WARNING: Using user-provided nvcc: 'nvcc -ccbin gcc'` when testing. I did not take a closer look at the autoconf macros but would assume they expand, don't they?

That warning is from line 82, where the single quote is inside a double-quoted string.
Line 88 is not an autoconf macro. It's normal shell variables. Since it works, I am probably missed something here. I'll follow up.

spickartz: > Line 88 is not an autoconf macro. It's normal shell variables. Since it works, I am probably missed something here. I'll follow up.

Yes, I know. But `ac_compile` is certainly used (and then expanded) by an autoconf macro I would have assumed.

spickartz: Fixed this and also changed the behavior w.r.t. `NVCC` and `NVCC_FLAGS`.

hzhou (approved):
Thanks for the changes!

hzhou:
test:yaksa/gpu

PR #241, opened 2023-01-31
fix libtool scripts for gpu backends

raffenet:
## Pull Request Description

Do not assume libtool is in PATH. This may not be the case, for instance
with preconfigured tarballs. Use the local copy.

## Expected Impact

Fix tarball builds when no `libtool` is available on the system.

## Author Checklist


raffenet:
Closing. Will approve #242.

PR #243, opened 2023-03-22
config/cuda: do not reset NVCC

hzhou:
## Pull Request Description
An extra check that resets NVCC.

Remove NVCC_FLAGS from makefile since it is already in the NVCC variable.

Also, move AC_SUBST to the outside as it does not depend on the branch.

Ref: https://github.com/pmodels/mpich/issues/6447

## Author Checklist


hzhou:
test:yaksa/gpu

raffenet (approved):

PR #246, opened 2023-09-16
Adding auto CUDA compute capability detection

yfguo:
## Pull Request Description

PR1 simply add options for newer CUDA devices
PR2 add the auto detection. For CUDA 11.5+, we just use the new `-arch=native` option. For older version, we need to run a program to query the device version.

`--with-cuda-sm=auto` is set as the new default. The behavior is also changed to the follow:
1. If `auto` is set, any else options are ignored. e.g. `--with-cuda-sm=70,auto,80 == --with-cuda-sm=auto`.
2. Auto detection will fail only when CUDA version < 11.5 and no GPU present in the building node. If auto detection is failed, ~~we does not revert back to `--with-cuda-sm=all`~~. Probably should revert back to build all and throw out a warning.
3. Other options kept the existing behavior

## Author Checklist

### DIFF
--- a/src/backend/cuda/subconfigure.m4
+++ b/src/backend/cuda/subconfigure.m4
@@ -49,6 +49,16 @@ AC_ARG_WITH([cuda-sm],
                 ampere - build compatibility for all Ampere GPUs
                 80     - A100, A30
                 86     - RTX Ampere, MX570, A40, A16, A10, A2
+                87     - Jetson AGX Orin and Drive AGX Orin
+
+                # Ada architecture
+                ada    - build compatibility for all Ada GPUs
+                89     - GeForce RTX 4090, RTX 4080, RTX 6000, Tesla L40
+
+                # Hopper architecture
+                hopper - build compatibility for all Hopper GPUs
+                90     - NVIDIA H100 (GH100)
+                90a    - add acceleration for features like wgmma and setmaxnreg. Required for NVIDIA CUTLASS
 
                 # Other
                 <numeric> - specific SM numeric to use
@@ -150,7 +160,7 @@ fi
 ##########################################################################
 
 if test "${have_cuda}" = "yes" ; then
-    for version in 11010 11000 10000 9000 8000 7000 6000 5000 ; do
+    for version in 12000 11080 11010 11000 10000 9000 8000 7000 6000 5000 ; do
         AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
                               #include <cuda.h>
                               int x[[CUDA_VERSION - $version]];
@@ -163,7 +173,13 @@ if test "${have_cuda}" = "yes" ; then
     for sm in ${with_cuda_sm} ; do
         case "$sm" in
             all)
-                if test ${cuda_version} -ge 11010 ; then
+                if test ${cuda_version} -ge 12000 ; then
+                    # maxwell (52) to hopper (90a)
+                    supported_cuda_sms="52 53 60 61 62 70 72 75 80 86 87 89 90 90a"
+                elif test ${cuda_version} -ge 11080 ; then
+                    # maxwell (52) to ada (89) and hopper (90)
+                    supported_cuda_sms="52 53 60 61 62 70 72 75 80 86 87 89 90"
+                elif test ${cuda_version} -ge 11010 ; then
                     # maxwell (52) to ampere (86)
                     supported_cuda_sms="52 53 60 61 62 70 72 75 80 86"
                 elif test ${cuda_version} -ge 11000 ; then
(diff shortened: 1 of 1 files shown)


yfguo:
OK. So `-arch=native` also does not work if there is no GPU on the build node.

Probably should check if `-arch=native` actually work and revert back to build all behavior.

src/backend/cuda/subconfigure.m4
```diff
@@ -169,6 +169,50 @@ if test "${have_cuda}" = "yes" ; then
...
     done
+    if test ${with_cuda_sm} = *auto* ; then
+        dnl process auto detection
+        PAC_PUSH_FLAG([IFS])
+        IFS=" "
+        AC_MSG_CHECKING([for CUDA compute capability auto detection])
+        AC_LANG_PUSH([C])
+        ac_ext=cu
+        ac_objext=out
+        ac_compile="$NVCC -o conftest.$ac_objext conftest.$ac_ext >&5"
```
raffenet, line 180: do we really need to use `nvcc` here or can we use the runtime API from C?

src/backend/cuda/subconfigure.m4
```diff
@@ -150,20 +162,74 @@ fi
...
+    else
+        with_cuda_sm=${$with_cuda_sm/auto/all}
+    fi
+
     PAC_PUSH_FLAG([IFS])
     IFS=","
     CUDA_SM=
     for sm in ${with_cuda_sm} ; do
         case "$sm" in
+            all_vert)
```
raffenet, line 221:
```suggestion
            all_virt)
```

src/backend/cuda/subconfigure.m4
yfguo: It is runtime API, let me try with C compiler.

src/backend/cuda/subconfigure.m4
yfguo: Will fix.

yfguo:
@raffenet I have also added the `all-major` option that Jeff wrote in #247.

src/backend/cuda/subconfigure.m4
```diff
@@ -160,16 +161,64 @@ fi
...
+                    cuda_output=$(./conftest$EXEEXT | xargs -n1 | sort -u | xargs)
+                    for sm in $cuda_output; do
+                        sm_no_decimal=`echo $sm | tr -d '.'`
+                        PAC_APPEND_FLAG([$sm_no_decimal],[CUDA_SM])
+                    done
+                    with_cuda_sm=
+                    AC_MSG_RESULT([yes])
+                ],
+                [
+                    with_cuda_sm=${$with_cuda_sm/auto/all}
```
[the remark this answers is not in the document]
yfguo: Yeah, this might be over complicated. I will change it to just `all`. (and `all-major` in later commits).

src/backend/cuda/subconfigure.m4
```diff
@@ -15,6 +15,7 @@ AC_ARG_WITH([cuda-sm],
           Comma-separated list of below options:
                 auto - automatically build compatibility for all GPUs visible, any other specified compatibilities are ignored
                 all - build compatibility for all GPUs supported by the CUDA version (can increase compilation time)
+                all-major - build compatibility for all major GPU versions (sm_*0) supported by the CUDA version
```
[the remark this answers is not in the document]
yfguo: Will do that. Do you think we should also remove `all`? I think `all-major` is a better fallback solution and I don't think anyone would wanted to build `all`.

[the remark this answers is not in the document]
raffenet: `all-major` as default works for me.

src/backend/cuda/subconfigure.m4
```diff
@@ -160,16 +161,64 @@ fi
...
+                      return 0;
+                  ]
+                )],
+                [
+                    cuda_output=$(./conftest$EXEEXT | xargs -n1 | sort -u | xargs)
+                    for sm in $cuda_output; do
+                        sm_no_decimal=`echo $sm | tr -d '.'`
+                        PAC_APPEND_FLAG([$sm_no_decimal],[CUDA_SM])
+                    done
+                    with_cuda_sm=
```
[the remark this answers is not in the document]
raffenet: Ah, I see how it works now. 👍

src/backend/cuda/subconfigure.m4
```diff
@@ -206,7 +206,7 @@ if test "${have_cuda}" = "yes" ; then
                     AC_MSG_RESULT([yes])
                 ],
                 [
-                    with_cuda_sm=${$with_cuda_sm/auto/all}
+                    with_cuda_sm=${$with_cuda_sm/auto/all-major}
```
raffenet, line 209: I'm still confused by this setting. What do the slashes do?

yfguo: Sorry, forgot to fix that in last push. Just pushed a new version

raffenet (approved):
LGTM assuming tests pass.

PR #248, opened 2023-10-03
Update CHANGES in preparation of 0.3 release

raffenet:
## Pull Request Description

There have been a good number of changes since 0.2. Since Spack defaults to using an external yaksa release to build MPICH, tagging 0.3 is a good idea so users can get the latest functionality.

## Author Checklist

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,32 @@
 ===============================================================================
+                               Changes in 0.3
+===============================================================================
+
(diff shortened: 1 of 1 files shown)


raffenet:
I dropped the commits fixing up the release script and version.m4 since what we have right now is working fine for Spack. We can address those items if we decide to do more formal releases, e.g. 1.0.

hzhou (approved):

PR #259, opened 2024-11-07
backend: Default nesting levels to 2

raffenet:
## Pull Request Description

This significantly reduces the build time and binary size, which is critical for large-scale.

## Expected Impact

Reduced binary size. Performance for deeply nested datatypes may suffer, but that is a secondary concern.

## Author Checklist


raffenet:
We already did this in MPICH. Just make it the default here for external builds e.g. in Spack.

hzhou (approved):
LGTM

PR #267, opened 2025-12-10
backend: Fix automated sm support for CUDA 13

natshineman:
## Pull Request Description

Updated the backend/cpu subconfigure to better support CUDA 13. Namely the all-major feature would try to build sm versions that are no longer supported by CUDA 13 resulting in compilation errors. Looking at the help output from nvcc, it looks like anything below sm_75 is fully removed from CUDA 13 and cannot be built at all. 

Also added a blackwell case for the arch naming sm version that uses 100/120(a) values to be inline with similar support for older architectures.

## Expected Impact
Support CUDA 13 via `--with-cuda-sm=all-major` and support `--with-cuda-sm=blackwell`. 

## Author Checklist


natshineman:
Closing so this can go to the MPICH repo

